PR #41, opened 2022-09-27
LOESS Regression

GeorgWa:
Feel free to move the module from `statistics.regression` to wherever it should reside.


alphabase/statistics/regression.py
```diff
@@ -0,0 +1,235 @@
...
+from sklearn.base import BaseEstimator, RegressorMixin
+
+from sklearn.utils.estimator_checks import check_estimator
+import matplotlib.pyplot as plt
+
+# %% ../../nbdev_nbs/statistics/regression.ipynb 3
+EPSILON = 1e-6
+
+class LOESSRegression(BaseEstimator, RegressorMixin):
+
```
jalew188, line 19: In nbdev2 (version last month), when we have doc_string below `__init__()`, nbdev2 will look for doc_string from base classses. It is annoying but we should follow its rule (or bug?). Here we can:

1. move global description of `LOESSRegression` right below the class declaration
2. keep parameters under `__init__()`

alphabase/statistics/regression.py
```diff
@@ -0,0 +1,235 @@
...
+__all__ = ['EPSILON', 'LOESSRegression']
+
+# %% ../../nbdev_nbs/statistics/regression.ipynb 2
+import numpy as np
+
+from sklearn.preprocessing import PolynomialFeatures
+from sklearn.base import BaseEstimator, RegressorMixin
+
+from sklearn.utils.estimator_checks import check_estimator
+import matplotlib.pyplot as plt
```
jalew188, line 13: matplotlib is not in requirements of alphabase, as it is only for testing, put it into a hide cell `#| hide ...`

alphabase/statistics/regression.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def _more_tags(self):
+        return {'X_types': ['1darray']}
+
+    def calculate_kernel_indices(self, x: np.ndarray):
+        """Determine the indices of the datapoints belonging to each kernel.
+
+        Parameters
+        -----------
+        x : numpy.ndarray, float, of shape (n_datapoints)
```
jalew188, line 58: For nbdev2, it is better to use:

```
x : np.ndarray
    float, of shape (n_datapoints)
```

Otherwise the contents after 'float' will be hided

jalew188:
Overall is very nice,  after move matplotlib import to a hidden cell, we can merge this PR

nbdev_nbs/statistics/regression.ipynb
```diff
@@ -0,0 +1,334 @@
+{
```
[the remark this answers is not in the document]
GeorgWa: There is still an issue if there fewer datapoints than `n_kernels * kernel_size`. Will change this in the future.

PR #44, opened 2022-10-12
updated weight function

GeorgWa:
Updated the weight function to allow for correct extrapolation even if all datapoints are outside the kernel limits.


jalew188:
Cool updates, I think we can use LOESS instead of KNN for different applications already.

jalew188 (approved):

PR #55, opened 2022-12-16
Reannotate precursors with fragments from different speclib

GeorgWa:
When re-extracting psms it can be necessary to annotate precursors with an existing spectral library. The `mod_seq_hash` column is used to left-joint fragments and non-matched precursors are removed. The whole `fragment_intensity_df` and `fragment_mz_df` are copied.

`SpecLibBase.annotate_fragments_from_speclib(SpecLibBase)`

`remove_unused_fragments` can be used after annotation or generally to remove fragments which are not used in the `precursor_df`. The operation happens in place and fragment start and end indices are recalculated.

`SpecLibBase.remove_unused_fragments()`


jalew188:
@GeorgWa Could you update the code in sphinx_doc branch? Otherwise there are quite many conflicts because I removed all `#| export` cells in nbdev_nbs

GeorgWa:
> @GeorgWa Could you update the code in sphinx_doc branch? Otherwise there are quite many conflicts because I removed all `#| export` cells in nbdev_nbs

Sure, will do 👍🏻

PR #84, opened 2023-01-26
use flat_frag_start/stop_idx other than frag_start/stop_idx for flatten fragments

jalew188:
keep both flat and unflat fragment pointers in case we need both of them in some applications


jalew188:
Other information:

1. flatten_fragments() becomes an inplace operation for precursor_df

2. All information were saved to hdf.library group, we will not use `flat_library` any more

GeorgWa:
I think this is a good idea! 
I only have a different question.
Before, we returned a copy of the dataframe which could then be assigned to the old variable. The classic pandas style:
`precursor_df = do_something(precursor_df)`

now, we do both, we change it in place and we return the updated dataframe.
Do you think this could be confusing? How do other functions handle this?

jalew188:
> There's only a different point: Before, we returned a copy of the dataframe which could then be assigned to the old variable. The classic pandas style: `precursor_df = do_something(precursor_df)`
> 
> now, we do both, we change it in place and we return the updated dataframe. Do you think this could be confusing? How do other functions handle this?

In AlphaBase, we have quite many `inplace+return` functions because copy of a big precursor_df is expensive (10GB for 10M/20M precursors), if we really want a copy, we can do

```
df = do_something(df).copy()
```

jalew188:
> **Edit** Sorry, just saw the message now. Now, this would also change the "original" speclib when creating a flat_speclib:
> 
> ```python
>     def parse_base_library(self, library:SpecLibBase):
>         """ Flatten an library object of SpecLibBase or its inherited class. 
>         This method will generate :attr:`precursor_df` and :attr:`fragment_df`
>         The fragments in fragment_df can be located by 
>         `frag_start_idx` and `frag_stop_idx` in precursor_df. 
> 
>         Parameters
>         ----------
>         library : SpecLibBase
>             A library object with attributes
>             `precursor_df`, `fragment_mz_df` and `fragment_intensity_df`.
>         """
>         self._precursor_df, self._fragment_df = flatten_fragments(
>             library.precursor_df, 
>             library.fragment_mz_df, 
>             library.fragment_intensity_df,
>             min_fragment_intensity=self.min_fragment_intensity,
>             keep_top_k_fragments=self.keep_top_k_fragments,
>             custom_columns=self.custom_fragment_df_columns,
>         )
> ```

`Now, this would also change the "original" speclib when creating a flat_speclib:` You mean we added two columns in the precursor_df?

GeorgWa:
Yep, in general this is a good point, but what about the `parse_base_library ` function?  There we will copy anyways, but we shouldn't change the original df?

I'm fine with both ways!

jalew188:
Maybe I misunderstood. 

I think the purpose of this PR is to avoid copy in `parse_base_library()`. After parsing, if we would like to call fragment_mz_df, we use the `frag_start_idx` column, otherwise using `flat_frag_start_idx`?

jalew188:
Good idea, I will add:

```
  def parse_base_library(self, 
      library:SpecLibBase,
      keep_original_frag_dfs:bool=False,
      copy_precursor_df:bool=False,
  ):
```

GeorgWa:
Sounds good! 👍🏻

jalew188:
```
def parse_base_library(self, 
        library:SpecLibBase,
        keep_original_frag_dfs:bool=False,
        copy_precursor_df:bool=False,
    ):
        self._precursor_df, self._fragment_df = flatten_fragments(
            library.precursor_df.copy() if copy_precursor_df else library.precursor_df, 
            library.fragment_mz_df, 
            library.fragment_intensity_df,
            min_fragment_intensity=self.min_fragment_intensity,
            keep_top_k_fragments=self.keep_top_k_fragments,
            custom_columns=self.custom_fragment_df_columns,
        )
        
        if hasattr(library, 'protein_df'):
            self._protein_df = library.protein_df
        else:
            self._protein_df = pd.DataFrame()

        if keep_original_frag_dfs:
            self.charged_frag_types = library.fragment_mz_df.columns.values
            self._fragment_mz_df = library.fragment_mz_df
            self._fragment_intensity_df = library.fragment_intensity_df
        else:
            self._fragment_mz_df = pd.DataFrame()
            self._fragment_intensity_df = pd.DataFrame()
```

jalew188:
@GeorgWa  Good suggestion, now this looks better.

GeorgWa (approved):

PR #86, opened 2023-02-10
remove is_pd_dataframe attr for DataFrame in hdf

jalew188:
The hdf can be opened by HDFView (v3.1.1)


jalew188:
Now the HDF file can be opened by HDFView (v3.1.1 tested), will fix issue https://github.com/MannLabs/alphapeptdeep/issues/90

PR #91, opened 2023-03-07
Updated Library Handling

GeorgWa:
**Changes:**
- renamed `SWATHLibraryReader` to `LibraryReaderBase`
- separate `library_reader_base` in `psm_reader.yaml` config for `LibraryReaderBase`
- New `LibraryReaderBase._get_fragment_intensity` implementation which is called at the end of the parsing process in `PSMReaderBase._post_process`. This allows it to operate only on the translated column names. By default, all non-fragment columns will be grouped and part of the final output.
- `SpecLibBase.copy()` for copying spectral libraries including all attributes.
- `SpecLibBase.append()` for appending spectral libraries while maintaining the fragment index mapping.


alphabase/peptide/fragment.py
```diff
@@ -1004,3 +1008,118 @@ def join_left(
...
+    """
+    
+    if len(precursor_df) == 0:
+        raise ValueError('Precursor dataframe is empty.')
+    
+    if len(fragment_mz_df) == 0:
+        raise ValueError('Fragment dataframe is empty.')
+    
+    if group_column not in precursor_df.columns:
+        raise ValueError('Group column not in precursor dataframe.')
```
jalew188, line 1049: `KeyError`

alphabase/psm_reader/maxquant_reader.py
```diff
@@ -101,6 +101,7 @@ def __init__(self,
         keep_decoy = False,
         fixed_C57 = True,
         mod_seq_columns = ['Modified sequence'],
+        csv_sep = '\t',
```
jalew188, line 104: In the later version, we determine csv_sep based on the given tsv file, so we should skip this param

alphabase/psm_reader/maxquant_reader.py
```diff
@@ -198,7 +200,7 @@ def _init_column_mapping(self):
         ]['column_mapping']
 
     def _load_file(self, filename):
-        df = pd.read_csv(filename, sep='\t')
+        df = pd.read_csv(filename, sep=self.csv_sep)
```
jalew188, line 203:
```
self.csv_sep = self._get_table_delimiter(filename)
df = pd.read_csv(filename, sep=self.csv_sep)
```

jalew188: However, MaxQuant's format is always TSV, we don't need sep param

alphabase/spectral_library/reader.py
```diff
@@ -20,22 +25,69 @@ def __init__(self,
...
+            False discovery rate threshold for filtering the spectral library.
+            default is 0.01
+
+        fixed_C57: bool
+    
+        mod_seq_columns: list of str
+            List of column names in the csv file containing the modified sequence.
+            By default the mapping is taken from `psm_reader.yaml`
+
+        csv_sep: str
```
jalew188, line 64: csv_sep is not needed anymore

alphabase/spectral_library/reader.py
```diff
@@ -160,28 +236,39 @@ def _get_fragment_intensity(self, lib_df:pd.DataFrame):
...
-
+        """
+        Process the spectral library and create the `fragment_intensity`, `fragment_mz`dataframe.
+        Reimplementation of `PSMReaderBase._post_process`.
+        """
+        
+        if 'nAA' not in self._psm_df.columns:
+            self._psm_df['nAA'] = self._psm_df.sequence.str.len()
+        
+        self._psm_df = self._get_fragment_intensity(self._psm_df)
```
jalew188, line 267: If some modifications are not in the AB mod list, the corresponding PSMs will be removed. It is better to test loading the lib with unknown modifications.

jalew188:
After this PR, I think we can upgrade to 1.0.3

alphabase/peptide/fragment.py
```diff
@@ -619,6 +620,9 @@ def flatten_fragments(
     else:
         use_intensity = False
 
+    for col_name, df in custom_df.items():
+        frag_df[col_name] = df.values.reshape(-1)
```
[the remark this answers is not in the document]
GeorgWa: `custom_df` is a dict of dataframes which have the same shape as `fragment_mz_df` and `fragment_intensity_df`. Therefore every dataframe will become a single column in the flattened representation. I've inculded an updated testcase and documentation to explain this: [833d21e](https://github.com/MannLabs/alphabase/pull/91/commits/833d21e08230caec4162349d010d21530e485e98)

alphabase/spectral_library/reader.py
GeorgWa: I don't quite know how to adress this issue, let's discuss it soon.

PR #99, opened 2023-05-10
add possibility to redefine path where tempmmap arrays are saved

sophiamaedler:
Currently there is no possibility to change the path of where memory mapped numpy arrays are written. I added a function to do this so that e.g. when you are deploying the io module on a cluster infrastructure you can define a path to large file storage system to use as a temp directory that is independent of where the code is being executed.


jalew188:
It looks very nice, thank you Sophia @sophiamaedler

PR #112, opened 2023-10-15
annotate precursor df with FASTA

GeorgWa:


GeorgWa:
@jalew188 There seems to be some error with the pfind reader which didn't occour with the previous stable tests.
This should not be affected by my changes at all, do you have any suspicion?

PR #130, opened 2024-01-13
Development

jalew188:
In spectral_library.base, `calc_precursor_isotope` will use `calc_precursor_isotope_intensity`. It will be the default option for all speclib generation if isotopes are included. 

I also removed `multiprocessing=True` parameter as it can be controled by `mp_process_num>1`. @GeorgWa Please be aware this change to alphadia.


alphabase/spectral_library/base.py
```diff
@@ -310,25 +310,25 @@ def clip_by_precursor_mz_(self):
...
     def calc_precursor_mz(self):
         """
-        Calculate precursor mz for self._precursor_df,
-        and clip the self._precursor_df using `self.clip_by_precursor_mz_`
+        Calculate precursor mz for self._precursor_df
         """
         fragment.update_precursor_mz(self._precursor_df)
 
-    def update_precursor_mz(self):
+    def calc_and_clip_precursor_mz(self):
```
GeorgWa, line 317: really good, I like that the clipping is mentioned and won't surprise users

alphabase/spectral_library/base.py
```diff
@@ -351,53 +351,70 @@ def calc_precursor_isotope_intensity(self,
...
                 self._precursor_df
             ) = precursor.calc_precursor_isotope_intensity(
                 self.precursor_df, 
                 max_isotope = max_isotope,
+                normalize=normalize,
                 min_right_most_intensity = min_right_most_intensity,
             )
-            
+
+    def calc_precursor_isotope(self,
```
GeorgWa, line 377: If you like we can keep only one `calc_precursor_isotope_intensity` or `calc_precursor_isotope` so there's no redundancy.

GeorgWa (approved):
Looks good for me!

alphabase/spectral_library/base.py
jalew188: `calc_precursor_isotope` is now `calc_precursor_isotope_info`. My original thought is that maybe it is useful in AlphaDIA, but I think your `calc_*_intensity` is better in readability and memory-saving.

jalew188:
@GeorgWa You this PR does not disturb some AlphaDIA functionalities, we can them merge it.

PR #138, opened 2024-03-04
SpecLlibFlat to specLibBase conversion

mo-sameh:


GeorgWa:
1. when inferring charged frag types, include all charges up to the maximum observed charge.
2. Move all test cases into a single notebook at spectral_library/flat_library.ipynb

Then we are ready to merge 🚀

PR #139, opened 2024-03-04
FIX Sage reader bug

GeorgWa:
I reported the `mod_sites` location for the modified sequence and not for the stripped sequence.


jalew188:
FIX https://github.com/MannLabs/alphapeptdeep/issues/143?

PR #142, opened 2024-03-13
rename diann spec_idx to diann_spec_idx

jalew188:
After pandas upgrade, some df.values will be readonly in Linux. This PR is designed to fix some read-only issues.


jalew188:
Fix #143 ?

GeorgWa (approved):

GeorgWa:
Hi Feng, looks good 👍 what is the origin of that bug? Is it Pandas?

Just to make sure that I understand it if it happens to me.

jalew188:
> Hi Feng, looks good 👍 what is the origin of that bug? Is it Pandas?
> 
> Just to make sure that I understand it if it happens to me.

Hi @GeorgWa, from the auto tests, I guess it is due to pandas version and only on Linux. I think all .values will be a readonly property instead of a ref of numpy array. I am not entirely sure anyway.

PR #145, opened 2024-03-16
FIX match only dense dataframes

GeorgWa:
`SpecLibBase.available_fragment_dfs()` would return `['_fragment_df', '_fragment_intensity_df', '_fragment_mz_df']` on a SpecLibFlat and fail on `.remove_unused_fragments()`.

This makes sure it's really matching `_fragment_[attribute_name]_df`.

### DIFF
--- a/alphabase/spectral_library/base.py
+++ b/alphabase/spectral_library/base.py
@@ -136,7 +136,7 @@ class SpecLibBase(object):
         return self._fragment_intensity_df
     
 
-    def available_fragment_dfs(self)->list:
+    def available_dense_fragment_dfs(self)->list:
         """
         Return the available dense fragment dataframes
         By dynamically checking the attributes of the object.
@@ -198,7 +198,7 @@ class SpecLibBase(object):
             
         """
         if remove_unused_dfs:
-            current_frag_dfs = self.available_fragment_dfs()
+            current_frag_dfs = self.available_dense_fragment_dfs()
             for attr in current_frag_dfs:
                 if attr not in dfs_to_append:
                     delattr(self, attr)
@@ -501,7 +501,7 @@ class SpecLibBase(object):
         Fragment dataframes are updated inplace and overwritten.
         """
 
-        available_fragments_df = self.available_fragment_dfs()
+        available_fragments_df = self.available_dense_fragment_dfs()
         non_zero_dfs = [
             df for df in available_fragments_df 
             if len(getattr(self, df)) > 0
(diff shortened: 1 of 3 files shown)


jalew188:
Looks nice

GeorgWa:
I discussed the topic a bit with @mo-sameh and we reasoned that there is currently some ambiguity between a `SpecLibBase` and `SpecLibFlat`. As the `SpecLibFlat` is inherited and produced by a `SpecLibBase` it can still have dense representations (`fragment_mz_df`, `fragment_intensity_df` etc.) and inherited functions like `remove_unused_fragments` which don't operate on the flat representation. This is confusing as one would expect they affect your flat library.

Therefore I made some small changes to this PR:

1. Renaming `available_fragment_dfs` to `available_dense_fragment_dfs` to make clear this only returns dense fragments matrices.
2. Returning an empty list on `SpecLibFlat.available_dense_fragment_dfs` to make clear that dense representations are not supported as part of a SpecLibFlat.
3. Get rid of dense representations when `SpecLibBase.parse_base_library` is called. If `keep_original_frag_dfs` is set to true, a deprecation warning is issued.
4. Raise a not implemented error if `SpecLibFlat.remove_unused_fragments` is called.

I think to handle complexity we have to keep the different library types better separated.

Looking forward to hear your thoughts

jalew188:
We should use SpecLibBase as the composite instead of the parent class.

mo-sameh:
The save/load_hdf functions in flat.py are still using the dense DataFrames. 
I agree that there's a need for refactoring the class structures and their relationships. Apart from that, for this PR everything looks good to me.

GeorgWa:
I agree, or we should have a base class which only handles precursor operations and have this as parent class for SpecLibBase and SpecLibFlat. I think this would be a separate bigger refactoring involving Magnus.

jalew188:
Yes, let's keep as it is now. You can merge I think

alphabase/spectral_library/base.py
```diff
@@ -135,9 +136,9 @@ def fragment_intensity_df(self)->pd.DataFrame:
         return self._fragment_intensity_df
     
 
-    def available_fragment_dfs(self)->list:
+    def available_dense_fragment_dfs(self)->list:
```
jalew188, line 139: What about move this into flat_lib? As a base lib must have dense frag dfs

GeorgWa: It's only applicable to the dense lib. The flat lib will always return an empty list.

The issue is that it's inherited.

PR #151, opened 2024-04-12
reconnect to existing HDF5 files using tempmmap

sophiamaedler:
In some cases it can be useful to reconnect to HDF5 files which still exist on the file system from previously utilised tempmmapped arrays. I added a function which implements this. All required parameters are read from the HDF5 file itself and only the path needs to be provided.


alphabase/io/tempmmap.py
```diff
@@ -90,6 +92,41 @@ def array(shape: tuple, dtype: np.dtype) -> np.ndarray:
             offset=offset
         ).reshape(shape)
 
+def map_array(hdf_file: str) -> np.ndarray:
```
jalew188, line 95: map_array looks like a dict/map array, which is ambiguious, maybe mmap_array is better.

jalew188:
I change the PR target branch to `development` to prevent conflicts of other PRs

jalew188:
Except for the function name, others look very good, I am actually going to use this explicit mmap

alphabase/io/tempmmap.py
sophiamaedler: Sure sounds look a good idea.

sophiamaedler:
Function name is updated. Should hopefully be good to go then. Thanks for the review @jalew188 .

alphabase/io/tempmmap.py
```diff
@@ -90,6 +92,41 @@ def array(shape: tuple, dtype: np.dtype) -> np.ndarray:
             offset=offset
         ).reshape(shape)
 
+def mmap_array(hdf_file: str) -> np.ndarray:
```
jalew188, line 95: Creating hdf5 file for mmap is missing, e.g., `save_array_for_mmap()`

jalew188: Otherwise users will not know how to create the hdf_file for mmap

sophiamaedler: Good Point. At the moment the HDF5 files are instantiated everytime a tempmmap is created. Whenever the process completes successfully and the array object is deleted the HDF5 file the results are mapped to are also deleted. 

 For me sometimes when a process does not complete properly (because it encounters an error before the temp directory is cleaned up) this file remains. Using the suggested function I can reconnect and continue on with what ever I was doing without needing to regenerate the array. 

How would you suggest  we should handle this? Since they are tempmmaps they are generally used for intermediate results which when computation is complete could be deleted. In some cases like the case outlined above it's useful to reconnect to one. But I am not sure how to make this behaviour more intuitive to a user.

jalew188: I see. We can merge this PR for our own usage. 

If we have a good idea to save the mmap hdf, we can create a new PR.

jalew188: I think we can explicitly create the mmap file, and delete it after our programs finish.

sophiamaedler: I went back and implemented an additional function to explicitly create the file for the mmap which can then be read by this function.  Can I append it to this or should I open a new PR?

jalew188: It is fine to append

sophiamaedler: I couldn't get it to append because it was already merged, sorry. But I made a new one and referenced this one.

PR #152, opened 2024-04-15
introduce explicit function for creating HDF5 for mapping to using mmap arrays.

sophiamaedler:
additional implementation of features discussed in #151

### DIFF
--- a/alphabase/io/tempmmap.py
+++ b/alphabase/io/tempmmap.py
@@ -92,7 +92,59 @@ def array(shape: tuple, dtype: np.dtype) -> np.ndarray:
             offset=offset
         ).reshape(shape)
 
-def mmap_array(hdf_file: str) -> np.ndarray:
+def create_empty_mmap(shape: tuple, dtype: np.dtype, path: str = None, overwrite = False):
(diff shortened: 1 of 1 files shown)


alphabase/io/tempmmap.py
```diff
@@ -92,7 +92,59 @@ def array(shape: tuple, dtype: np.dtype) -> np.ndarray:
...
+        array = hdf_file.create_dataset(
+            "array",
+            shape=shape,
+            dtype=dtype
+        )
+        array[0] = 0
+
+    return temp_file_name
+
+def mmap_array_from_path(hdf_file: str) -> np.ndarray:
```
jalew188, line 147: Yes, this name is better than mmap_array

sophiamaedler: agree :) 
I am currently using it to perform multi-threaded operations that write results to the same mmap object concurrently and its working perfectly. Interestingly, using this setup I don't encounter issues where the HDF5 file becomes locked when I have a writable instance as I do using base h5py.

jalew188: I think HDF5 locks is required for SLURM, multi-threading (not multi-processing) is fine

PR #157, opened 2024-04-17
156 change which tests are run in ci pipeline

mschwoer:
* Run full test matrix only on merge to main to save resources,  running a subset (ubuntu, loose) on every PR.
 * Add reusable workflow for running tests to avoid code duplication.

* Adds all working "doc" and "nbs_tests" notebooks to the tests in order to maximize their scope.

* Also, I fixed two obvious bugs in notebooks

Note that this should just be a temporary state: we should move all nbdev tests to pytest (but still run the documentation notebooks as part of CI)


nbs_tests/test_thread.ipynb
```diff
@@ -34,7 +34,7 @@
    "metadata": {},
```
mschwoer, line 34: <pre class="ql-syntax" spellcheck="false">Line #1.    from multiprocessing.pool import ThreadPool
</pre><p>just a test what happens to these comments</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MannLabs/alphabase/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

.github/workflows/_run_tests.yml
```diff
@@ -0,0 +1,61 @@
...
+        conda activate alphabase
+        INCLUDED_NBS=$(find ./docs/nbs -name "*.ipynb" | grep -v tutorial_dev_spectral_libraries.ipynb)
+        python -m pytest --nbmake $(echo $INCLUDED_NBS) 
+        conda deactivate
+    - name: Test additional notebooks
+      shell: bash -l {0}
+      # TODO make test_isotope_mp.ipynb work
+      run: |
+        conda activate alphabase
+        INCLUDED_NBS=$(find ./nbs_tests -name "*.ipynb" | grep -v test_isotope_mp.ipynb)
```
jalew188, line 59: In my previous tests, multiprocessing in ipynb sometimes suspended on some versions of Windows? That's why I moved some nbs into `nbs_tests` folder to prevent auto-testing. Will nbmake solve this issue?

mschwoer: ah, I was already wondering about the reasoning for this .. thank you for pointing me at this.
It could well be that under the hood nbmake does the same as nbdev, and thus also has issues with multiprocessing.
I would nevertheless give it a try. I added a comment what to do in that case.

GeorgWa (approved):
LGTM!

.github/workflows/_run_tests.yml
jalew188: Yes, if tests halted in the future, we know where to fix.

.github/workflows/pip_installation.yml
```diff
@@ -1,64 +1,39 @@
...
+
+# For feature branches, we don't test the full matrix (os x [stable, loose]) in order to save time & resources.
+# On each push to develop, the whole test matrix runs, additionally, its success is mandatory for merging to main.
+# In the (assumed to be rare case) that these full matrix tests fail but the reduced matrix tests passed,
+# we will learn about it latest before pushing to main. The merge that caused the issue can then be identified
+# from the tests that ran on develop.
 on:
   push:
-    branches: [ main ]
+    branches: [ main, develop ]
```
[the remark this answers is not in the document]
GeorgWa: The idea would be to run the full version list only after the PR merge. It will still run `ubuntu loose` on every PR commit.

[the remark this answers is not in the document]
mschwoer: thanks @jalew188 for spotting this :-)

.github/workflows/pip_installation.yml
```diff
@@ -2,63 +2,31 @@ on:
   push:
     branches: [ main ]
```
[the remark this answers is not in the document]
mschwoer: added "development" for this push rule. As discussed in person: let's stick with the reduced tests on branches for now, and see if we need to adjust in the future

jalew188 (approved):

PR #162, opened 2024-05-06
add functionality to specify temp directory location when creating memory mapped temp arrays

sophiamaedler:
this allows you to also utilize memory mapped temp arrays in multithreaded processes by explicitly passing the directory name without needing to rely on the global TEMP_DIR_NAME variable still being accurate


alphabase/io/tempmmap.py
```diff
@@ -104,12 +121,23 @@ def create_empty_mmap(shape: tuple, dtype: np.dtype, path: str = None, overwrite
     overwrite : bool , optional
         If True the file will be overwritten if it already exists.
         Defaults to False.
+    tmp_dir_name : str, optional
```
mschwoer, line 124: maybe rename to `temp_dir_abs_path`, and also explain that an absolute path is expected?

jalew188:
@sophiamaedler We now have quite restrict PR rules for alphaX packages. Please `git pull` before coding/pushing:)

alphabase/io/tempmmap.py
```diff
@@ -65,12 +65,23 @@ def array(shape: tuple, dtype: np.dtype) -> np.ndarray:
...
         The np.dtype of the array.
+    tmp_dir_name : str, optional
+        If specified the memory mapped array will be created in this directory. Defaults to None.
 
     Returns
     -------
     type
         A writable temporary mmapped array.
     """
+    # redefine the temporary directory if a new location is given otherwise read from global variable
```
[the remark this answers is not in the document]
jalew188: Yes. It had randint before, but I guess we modified it some time ago to reuse the folder. Now, using temp_dir_name, we can actually decide if we need to reuse the mmap.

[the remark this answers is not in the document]
GeorgWa: I think we should definetly aim for this!
I assumed the names of the mmap blobs themselve are already randomized?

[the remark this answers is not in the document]
jalew188: @GeorgWa ahh, good point.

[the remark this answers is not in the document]
sophiamaedler: The folder names are still randomised. This is the default when producing a TemporaryDirectory using tempfile. 

Screenshot 2024-05-08 at 10 53 11

I am not sure why the file lock becomes an issue with multiple threads on linux. According to the documentation: The file is readable and writable only by the creating user ID. I was thinking that potentially the generated temp directories are only accessible by the thread that created them? Maybe someone more well-versed in Linux file permissions/thread operations would be able to give input on that.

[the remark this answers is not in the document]
sophiamaedler: In general I think it would be good to find a robust solution that works across platforms and threads . I use the memory mapped temp arrays quite heavily in SPARCSpy and I guess I have a couple of unique use-cases.

1. Creating an empty memory mapped array that different processes can reconnect to (e.g. I have 4 channel images, each process modifies the information in one channel and then writes the results back to  the channel slot in the same memory mapped temp array)
2. individual processes have their own temp directories that are cleared as soon as they complete (this way we minimise disk usage as no longer needed things are cleared up straight away instead of at the end of the whole process)

The fixes with creating an empty memory mapped array and reconnecting to it (#151 and #152) addressed point 1. For point 2 I needed to be able to pass the specific path in which to produce a memory mapped temp array to ensure that it worked across threads on linux (#162). 

The implementation itself is great because you can use them just like numpy arrays but easily circumvent memory limitations. I have also found them to be fast and robust. So it would be great if we find a cleaner solution that maybe makes them more widely applicable.

sophiamaedler:
> @sophiamaedler We now have quite restrict PR rules for alphaX packages. Please `git pull` before coding/pushing:)

@jalew188 Sure :) I had done a git pull before opening this PR, should I still do anything else next time?

alphabase/io/tempmmap.py
```diff
@@ -104,12 +121,23 @@ def create_empty_mmap(shape: tuple, dtype: np.dtype, path: str = None, overwrite
...
+    # this allows you to ensure that the correct temp directory location is used when working with multiple threads
+    if tmp_dir_name is not None:
+        if os.path.isdir(tmp_dir_name):  # ensure that directory exists
+            TEMP_DIR_NAME = tmp_dir_name
+        else:
+            raise ValueError(
+                "The directory in which the file should be created does not exist."
+            )
```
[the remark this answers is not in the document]
sophiamaedler: Would I need to return the variable name as otherwise TEMP_DIR_NAME is only updated within the scope of the function and outside of the function we are still using the previously defined global variable?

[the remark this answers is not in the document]
sophiamaedler: I wrote a function which returns the path name if it fulfills the conditions and otherwise raises a warning. This should let us keep the same functionality without having to use global variables.

alphabase/io/tempmmap.py
```diff
@@ -51,6 +51,44 @@ def _check_temp_dir_existence(abs_path: str) -> str:
             f"The directory {abs_path} in which the file should be created does not exist."
         )
 
+def _check_file_location(abs_file_path: str, overwrite = False) -> str:
```
sophiamaedler, line 54: similar to your feedback in the review I also took the opportunity to move the other longer if else block to a separate function for better readability. @mschwoer let me know what you think.

mschwoer: +1, good idea!
just a small nitpicking on the name: from a method that is named `check_..` I would not expect a return value .. mb rename to `_get_file_location` ?

alphabase/io/tempmmap.py
```diff
@@ -22,6 +22,73 @@
...
+    Returns
+    ------
+    str 
+        The directory path if it exists.
+    """
+    #ensure that the path exists
+    if os.path.exists(abs_path):
+        # ensure that the path points to a directory
+        if os.path.isdir(abs_path):  
+            TEMP_DIR_NAME = abs_path
```
mschwoer, line 43: I think this is a decisive side effect that should be reflected in both the method name and the docstring ;-)

jalew188:
> > @sophiamaedler We now have quite restrict PR rules for alphaX packages. Please `git pull` before coding/pushing:)
> 
> @jalew188 Sure :) I had done a git pull before opening this PR, should I still do anything else next time?

I think the only thing is to fix this error in this PR: https://github.com/MannLabs/alphabase/actions/runs/9000693343/job/24725402319?pr=162#step:4:246

alphabase/io/tempmmap.py
sophiamaedler: That's true. I chose check because we already provide the file location before it's just a sanity check to make sure everything is as expected before proceeding. Happy to change though if you think its better understandable!

sophiamaedler:
> > > @sophiamaedler We now have quite restrict PR rules for alphaX packages. Please `git pull` before coding/pushing:)
> > 
> > 
> > @jalew188 Sure :) I had done a git pull before opening this PR, should I still do anything else next time?
> 
> I think the only thing is to fix this error in this PR: https://github.com/MannLabs/alphabase/actions/runs/9000693343/job/24725402319?pr=162#step:4:246

I installed the ruff linter and let it run over the code. I hope that should fix it and should avoid this problem for the future.

alphabase/io/tempmmap.py
mschwoer: e.g. rename to `_set_new_temp_dir`

alphabase/io/tempmmap.py
```diff
@@ -97,40 +180,33 @@ def create_empty_mmap(shape: tuple, dtype: np.dtype, path: str = None, overwrite
...
-                )
-        if not os.path.basename.endswith(".hdf"):
-            raise ValueError("The chosen file name needs to end with .hdf")
-        if os.path.isdir(os.path.commonpath(path)):
-            temp_file_name = path
-        else:
-            raise ValueError(
-                "The directory in which the file should be created does not exist."
-            )
+        temp_file_name = _check_file_location(file_path, overwrite=False)
```
mschwoer, line 209: e.g. rename to `_get_temp_file_ab_path`

alphabase/io/tempmmap.py
sophiamaedler: I don't think I understand what you mean with decisive side effect? This function does not actually do anything except raise a value error if the provided path does not fulfil the requirements (here that it exists and that it points to a directory).

jalew188 (approved):
tempmmap was designed long time ago by Sander, we can refactor it if necessary in the near future. 

The same for alphabase.io.hdf.py and so on.

alphabase/io/tempmmap.py
mschwoer: it does something very fundamental: it changes `TEMP_DIR_NAME`, thus altering the state of the module (which is exactly what we want at this point, so all fine, just the name should reflect that)

alphabase/io/tempmmap.py
sophiamaedler: renamed to `_get_file_location`.

sophiamaedler:
I also noticed that the global definition of TEMP_DIR_NAME was removed and added it back in. This otherwise throws errors in some rare cases.

sophiamaedler:
> tempmmap was designed long time ago by Sander, we can refactor it if necessary in the near future.
> 
> The same for alphabase.io.hdf.py and so on.

In general I think there is a lot of potential here for additional use cases e.g. I have currently been playing around with Dask arrays and found a way to use a tempmmap compatible HDF5 in the backend of a memory mapped Dask array (see here : [src/sparcstools/base/daskmmap.py](https://github.com/MannLabs/SPARCStools/blob/code_refactoring/src/sparcstools/base/daskmmap.py)) but I am not sure where the best place to include it would be and if alpha base is the correct location to add something like that. Might be worth a more in depth discussion at some point.

alphabase/io/tempmmap.py
```diff
@@ -65,12 +132,24 @@ def array(shape: tuple, dtype: np.dtype) -> np.ndarray:
...
     type
         A writable temporary mmapped array.
     """
+
+    global TEMP_DIR_NAME
+
+    # redefine the temporary directory if a new location is given otherwise read from global variable
+    # this allows you to ensure that the correct temp directory location is used when working with multiple threads
+    if tmp_dir_abs_path is not None:
+        TEMP_DIR_NAME = _change_temp_dir_location(tmp_dir_abs_path)
```
mschwoer, line 151: do we need this assignment here? after all, `TEMP_DIR_NAME` is set in `_change_temp_dir_location()` .. either remove the assignment here (and move the `global` to `_change_temp_dir_location` ?

mschwoer (approved):
LGTM! thanks for hanging on ;-)

PR #163, opened 2024-05-15
Remove warning log on module load

mschwoer:
Move the warning that was issues on importing the `tempmmap` module to the first call of the associated methods, to avoid effects like this:
image

@jalew188 we could think implementing a similar mechanism for creating the actual `TemporaryDirectory` object, as importing this module has the side effect of always creating an (empty) temp folder.


GeorgWa (approved):
I think this combines the best of both worlds. The warning is still issued but not on import (like when calling alphadia -h).

jalew188:
It shouldn't be warning I think. I think we can just logging.info or print(...).

Looks good to me

jalew188 (approved):

mschwoer:
waiting for https://github.com/MannLabs/alphabase/pull/162 to be merged to avoid @sophiamaedler having to deal with potential merge conflict :-)

PR #170, opened 2024-05-17
revert back to run reduced set of tests on branches

mschwoer:
as discussed in person: the branch protection on develop allows us to restrict tests on branches

### DIFF
--- a/.github/workflows/pip_installation.yml
+++ b/.github/workflows/pip_installation.yml
@@ -9,7 +9,7 @@ on:
   push:
     branches: [ main, development ]
   pull_request:
-    branches: [ main, development ]
+    branches: [ main ]
   workflow_dispatch:
 
 name: Pip installation and tests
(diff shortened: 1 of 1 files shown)


.github/workflows/pip_installation.yml
```diff
@@ -9,7 +9,7 @@ on:
   push:
     branches: [ main, development ]
   pull_request:
-    branches: [ main, development ]
+    branches: [ main ]
```
jalew188, line 12: We don't need the tests on PR merging into development?

mschwoer: I would suggest running tests only on 1 os,, cf. https://github.com/MannLabs/alphabase/blob/284b2f6241a214247f39a446b3e0f330367a8047/.github/workflows/pip_installation.yml#L3C1-L6C1

jalew188: I see.

jalew188 (approved):

PR #171, opened 2024-05-19
Remove noqa

jalew188:
- Fix noqa: modify imports of `alphabase.constants.element` to `alphabase.constants.atom`; add FutureWarning.
- Fix noqa: modify imports of `alphabase.peptide.precursor.is_precursor_sorted()` to `alphabase.peptide.precursor.is_precursor_refined()`; add FutureWarning.
- Fix noqa: modify imports of `alphabase.peptide.precursor.calc_precursor_mz()` to `alphabase.peptide.precursor.update_precursor_mz()`, and add FutureWarning.


alphabase/peptide/precursor.py
```diff
@@ -126,7 +135,16 @@ def update_precursor_mz(
     return precursor_df
```
GeorgWa, line 135: Do you think we should return None to emphasize that this function is inplace? 

I think it's a bit dangerous if you are inplace and still return a dataframe as this pattern is used in pandas for not-inplace operations.

GeorgWa (approved):
Looks good to me!

alphabase/peptide/precursor.py
jalew188: Yes. I have been thinking of this for a long time, I think we should learn pytorch that inplace methods are `xxxx_()` and return None.

jalew188: We shall create a ticket about this in alphaX development projects.

GeorgWa: I think if we agree on the naming and document this in the design philosophy its fine. I like the xxx_() and return None pattern.

alphabase/spectral_library/base.py
```diff
@@ -334,7 +334,7 @@ def calc_precursor_mz(self):
         """
         Calculate precursor mz for self._precursor_df
         """
-        fragment.update_precursor_mz(self._precursor_df)
+        precursor.update_precursor_mz(self._precursor_df)
```
mschwoer, line 337: we could also import `update_precursor_mz()` directly from `precursor`

mschwoer (approved):
LGTM! I like the approach using FutureWarnings

Issue #181, opened 2024-06-04
ERROR: 'charmap' codec can't decode byte 0x81 in position 4429...


hcadre:
**Describe the bug**
Running alphaDIA (one-Click installer for Windows) with the example HeLa test samples and the parameters as described resulted in this error thrown after a few seconds:

![image](https://github.com/MannLabs/alphadia/assets/16349515/f62432cb-20c2-41b7-83d5-30bbb5556613)

Any idea about what went wrong?

GeorgWa:
@hcadre thank you for uploading the fasta. It's not malformed, alphaBase just tries to parse it as utf8 and fails doing so. I have transferred the issue and will close it as soon as we have a fix.

LoayJabre:
Hi - 
I tried re-running everything today, and it worked fine! 

I tried with and without the .fasta and it worked. It must have been something I was doing wrong initially. 

Out of curiosity, is there a column key anywhere for the precursors.tsv file that is outputted after the search? I'm fairly new to this, and it would help a lot if there is a key to help me understand the output a bit better.

Thanks so much for being responsive! 

Loay

GeorgWa:
Hi, that's great! 
We have also made some change to prevent this error in the future.

As alphaDIA is still in beta and there are lot's of breaking changes we haven't yet done this.
I will try to do this in the next days 👍🏻

PR #182, opened 2024-06-13
Mitigate wrong encoding under windows

GeorgWa:
I'm not a 100% sure about the bug but I think python applies the wrong encoding in some versions of windows or locales. 
Based on the stack trace it seems like it tries to decode a file with cp1252 encoding and fails doing so.

This change forces utf8 encoding by default which is compatible with ascii.

### DIFF
--- a/alphabase/protein/fasta.py
+++ b/alphabase/protein/fasta.py
@@ -46,7 +46,7 @@ def read_fasta_file(fasta_filename: str = ""):
         protein information,
         {protein_id:str, full_name:str, gene_name:str, description:str, sequence:str}
     """
-    with open(fasta_filename, "rt") as handle:
+    with open(fasta_filename, "rt", encoding="utf-8", errors="ignore") as handle:
         iterator = SeqIO.parse(handle, "fasta")
         while iterator:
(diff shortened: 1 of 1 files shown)


jalew188 (approved):
LGTM! It is better to test by user's fasta.

alphabase/protein/fasta.py
```diff
@@ -46,7 +46,7 @@ def read_fasta_file(fasta_filename: str = ""):
         protein information,
         {protein_id:str, full_name:str, gene_name:str, description:str, sequence:str}
     """
-    with open(fasta_filename, "rt") as handle:
+    with open(fasta_filename, "rt", encoding="utf-8", errors="ignore") as handle:
```
mschwoer, line 49: should we really ignore errors here?

GeorgWa:
I think we can remove the `errors='ignore'` to be on the safe side.
Regarding testing, I can't reproduce the bug because the file in #181 itself is a valid utf-8.

jalew188:
I think it is the windows issue then, as different Windows in different languages open files in different codes.

PR #189, opened 2024-06-27
windowsfix for np.int64

machth:
changed two lines of code for windows fix to set standard dtype of np.int64, which has been a float on windows


jalew188 (approved):
This looks more definitive

sophiamaedler:
the issue with the testing loose pip installation seems to be a general issue in the development branch correct? How should we proceed so that we can merge this PR to support the memory mapped temp arrays also on windows? Anything that still needs to be done within this PR?

jalew188:
@sophiamaedler  Done

PR #196, opened 2024-07-06
load modseq by default

GeorgWa:
This should be a stimulus for a discussion:

The original speclib base implementation aimed to be high performant and only included key numeric columns.
If any other column should be loaded or such the `load_mod_seq` key needs to be set.

This is nice but very counterintuitive. I myself have encountered multiple times where my speclib "lost" columns and I had to remember to set this flag, which has now become my default. I looked into this and I have 29 occurences where I set this in my private notebooks and 6 occurences in alphaDIA :D.

Screenshot 2024-07-06 at 10 50 44

I decided to make this PR because @boopthesnoot has also steopped into this trap but solved it in a different way:
```python
self.speclib_flat.key_numeric_columns.extend(self.speclib_flat.precursor_df.columns.to_list())
```

I Propose to:
1. Change the default behaviour to the slow but user friendly case of loading all columns (this PR).
2. Remove the concept of `key_numeric_columns` and load all columns all of the time.

### DIFF
--- a/alphabase/spectral_library/base.py
+++ b/alphabase/spectral_library/base.py
@@ -626,7 +626,7 @@ class SpecLibBase(object):
             "fragment_intensity_df": self._fragment_intensity_df,
         }
 
-    def load_hdf(self, hdf_file: str, load_mod_seq: bool = False):
+    def load_hdf(self, hdf_file: str, load_mod_seq: bool = True):
         """Load the hdf library from hdf_file
 
         Parameters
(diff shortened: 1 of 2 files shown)


alphabase/spectral_library/base.py
```diff
@@ -626,7 +626,7 @@ def save_hdf(self, hdf_file: str):
             "fragment_intensity_df": self._fragment_intensity_df,
         }
 
-    def load_hdf(self, hdf_file: str, load_mod_seq: bool = False):
+    def load_hdf(self, hdf_file: str, load_mod_seq: bool = True):
```
jalew188, line 629: Please modify the docstring value as well

jalew188 (approved):
Make sense

boopthesnoot (approved):

PR #199, opened 2024-07-16
dynamic generation of SMILES for PTMs

boopthesnoot:
You can find the description and examples in `docs/nbs/tutorial_smiles.ipynb`


GeorgWa:
I would suggest to integrate the data with the modification.tsv and the amino_acids.yaml:
https://github.com/MannLabs/alphabase/blob/737f25c79c34de5182140543bf887ab61d7e53d5/alphabase/constants/const_files/amino_acid.yaml

So we don't have to add a new constants folder and need to have double book keeping of modification names. I think we can add two keys for each entry: `sum/composition` and `smiles`.

In the modifications.tsv you can just add a `smiles` column.

extra_requirements/development.txt
```diff
@@ -32,3 +32,4 @@ pyahocorasick
 pytest
 pre-commit==3.7.0
 nbmake==1.5.3
+rdkit
```
mschwoer, line 35: you use rdkit in production code, right? should be added to the other requirements.txt (please also use a fixed version)

tests/run_tests.sh
```diff
@@ -9,3 +9,5 @@ python -m pytest --nbmake $(echo $INCLUDED_NBS)
 
 INCLUDED_NBS=$(find ../nbs_tests -name "*.ipynb")
 python -m pytest --nbmake $(echo $INCLUDED_NBS)
+
+python -m pytest test_smiles.py
```
mschwoer, line 13: I think `python -m pytest` will suffice, it discovers files that contain tests itself..

alphabase/smiles/smiles.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Args:
+        aa_smiles (str): SMILES of an amino acid with or without PTMs, has to have the placeholder atoms [Xe] for N-term and [Rn] for C-term.
+        n_term_mod (Optional[str]): N-terminal modification name. Options are the keys in the n_term_modifications dict.
+        c_term_mod (Optional[str]): C-terminal modification name. Options are the keys in the c_term_modifications dict.
+
+    Raises:
+        ValueError: If the amino acid is invalid, or if N-terminal / C-terminal modification is not recognized.
+    """
+    if N_TERM_PLACEHOLDER not in aa_smiles or C_TERM_PLACEHOLDER not in aa_smiles:
```
mschwoer, line 36: could it be that multiple of these errors happen at once? if so, consider this pattern:
```
errors =[]
if N_TERM_PLACEHOLDER not in aa_smiles or C_TERM_PLACEHOLDER not in aa_smiles:
  errors.append(f"The amino acid {aa_smiles} must contain the N-terminal and C-terminal placeholders in order to be modified.")
if ...
if ....

if errors:
  ValueError("\n".join(errors))

```

This way, the user gets all errors at once, and doesnt have to play "error ping pong" :-)

docs/nbs/tutorial_smiles.ipynb
```diff
@@ -0,0 +1,355 @@
+{
```
mschwoer, line 1: very cool notebook!

tests/test_smiles.py
```diff
@@ -0,0 +1,116 @@
...
+
+def test_invalid_amino_acid_smiles():
+    with pytest.raises(ValueError):
+        modify_amino_acid("InvalidSMILES")
+
+
+def test_dimethyl_n_term_modification(alanine_smiles):
+    result = modify_amino_acid(alanine_smiles, n_term_mod="Dimethyl")
+    expected = "CN(C)[C@@H](C)C(=O)O"
+    assert Chem.MolToSmiles(Chem.MolFromSmiles(result)) == Chem.MolToSmiles(
```
mschwoer, line 102: could we not just do `assert result == expected` here?

GeorgWa:
@boopthesnoot Have a look at #200.
I created an extra and a decorator for rdkit.

boopthesnoot:
> I would suggest to integrate the data with the modification.tsv and the amino_acids.yaml: https://github.com/MannLabs/alphabase/blob/737f25c79c34de5182140543bf887ab61d7e53d5/alphabase/constants/const_files/amino_acid.yaml
> 
> So we don't have to add a new constants folder and need to have double book keeping of modification names. I think we can add two keys for each entry: `sum/composition` and `smiles`.
> 
> In the modifications.tsv you can just add a `smiles` column.

@GeorgWa But the smiles in the modifications.tsv will be a mess, some of them will be AA's with PTMs, some of them will be terminal modifications only, without the AA, and we would still have to store which is which somewhere. 
By adding a key for each of the AAs in amino_acids.yaml we'll still have double bookkeeping of the atomic composition because we can infer it from SMILES. Ofc it would mean having a rdkit dependency for the whole package x)

GeorgWa:
> @GeorgWa But the smiles in the modifications.tsv will be a mess, some of them will be AA's with PTMs, some of them will be terminal modifications only, without the AA, and we would still have to store which is which somewhere. By adding a key for each of the AAs in amino_acids.yaml we'll still have double bookkeeping of the atomic composition because we can infer it from SMILES. Ofc it would mean having a rdkit dependency for the whole package x)

We could resolve this by looking up the localizer `@Any N-Term`. Alternatively we can also introduce a second column `location = {'N','C','AA'}` which would use dynamic or fixed smiles depending of the value. 

In alphabase the modification names like`Dimethyl@K` are the primary keys across all applications. I think this primary key should only be defined once. Furthermore, the master record in `modifications.tsv` is updated automatically from unimod if more modifications are added. This way everything will stay in sync.

jalew188:
> > @GeorgWa But the smiles in the modifications.tsv will be a mess, some of them will be AA's with PTMs, some of them will be terminal modifications only, without the AA, and we would still have to store which is which somewhere. By adding a key for each of the AAs in amino_acids.yaml we'll still have double bookkeeping of the atomic composition because we can infer it from SMILES. Ofc it would mean having a rdkit dependency for the whole package x)
> 
> We could resolve this by looking up the localizer `@Any N-Term`. Alternatively we can also introduce a second column `location = {'N','C','AA'}` which would use dynamic or fixed smiles depending of the value.
> 
> In alphabase the modification names like`Dimethyl@K` are the primary keys across all applications. I think this primary key should only be defined once. Furthermore, the master record in `modifications.tsv` is updated automatically from unimod if more modifications are added. This way everything will stay in sync.

Yes, I think we should use only one PTM and AA defination file to remove ambiguity in the future.

extra_requirements/development.txt
jalew188: Agree!

tests/test_smiles.py
boopthesnoot: This makes the tests agnostic to the logic of generating the smiles string, this way if they for some reason change from what atom they start, we'll be fine

alphabase/smiles/smiles.py
boopthesnoot: Fixed :)

tests/run_tests.sh
boopthesnoot: thank you, fixed

extra_requirements/development.txt
boopthesnoot: added

alphabase/constants/aa.py
```diff
@@ -53,13 +58,11 @@ def reset_AA_df():
     AA_DF = pd.DataFrame()
     AA_DF["aa"] = [chr(aa) for aa in range(len(AA_ASCII_MASS))]
     AA_DF["formula"] = [""] * len(AA_ASCII_MASS)
-    aa_idxes = []
-    formulas = []
-    for aa, formula in AA_Formula.items():
-        aa_idxes.append(ord(aa))
-        formulas.append(formula)
-    AA_DF.loc[aa_idxes, "formula"] = formulas
+    AA_DF["smiles"] = [""] * len(AA_ASCII_MASS)
```
mschwoer, line 61: could you introduce `num_rows =  len(AA_ASCII_MASS)` and use it throught this method?

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
...
+
+
+class ChemicalCompositonFormula:
+    def __init__(self, formula):
+        self.elements = self._parse_formula(formula)
+
+    @classmethod
+    def from_rdkit_mol(cls, mol: Chem.Mol):
+        if not mol:
+            raise ValueError("Invalid RDKit molecule")
```
mschwoer, line 216:
```suggestion
            raise ValueError(f"Invalid RDKit molecule: {mol}")
```

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
         mass of the formula
     """
     return np.sum([CHEM_MONO_MASS[elem] * n for elem, n in parse_formula(formula)])
+
+
+class ChemicalCompositonFormula:
```
mschwoer, line 209: please add docstrings to the class and all methods

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
...
+        return cls._from_rdkit_formula(formula)
+
+    @classmethod
+    def _from_rdkit_formula(cls, formula):
+        instance = cls.__new__(cls)
+        instance.elements = instance._parse_rdkit_formula(formula)
+        return instance
+
+    def _parse_formula(self, formula):
+        pattern = r"(\d+)?([A-Z][a-z]*)(?:\(([-]?\d+)\))?"
```
mschwoer, line 230: could you add a comment what this regexp does?

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
...
+        for isotope, element, count in matches:
+            if isotope:
+                element = f"{isotope}{element}"
+            count = int(count) if count else 1
+            element_counts[element] += count
+
+        return element_counts
+
+    def _parse_rdkit_formula(self, formula):
+        pattern = r"(\[(\d+)([A-Z][a-z]*)\]|([A-Z][a-z]*))(\d*)"
```
mschwoer, line 243: could you add a comment what this regexp does?

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
...
+        matches = re.findall(pattern, formula)
+        element_counts = defaultdict(int)
+
+        for match in matches:
+            if match[1]:  # Isotope
+                element = f"{match[1]}{match[2]}"
+                count = int(match[4]) if match[4] else 1
+            else:  # Regular element
+                element = match[3]
+                count = int(match[4]) if match[4] else 1
```
mschwoer, line 253: the `count = int(match[4]) if match[4] else 1` is duplicated

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
...
+
+        return element_counts
+
+    def _parse_rdkit_formula(self, formula):
+        pattern = r"(\[(\d+)([A-Z][a-z]*)\]|([A-Z][a-z]*))(\d*)"
+        matches = re.findall(pattern, formula)
+        element_counts = defaultdict(int)
+
+        for match in matches:
+            if match[1]:  # Isotope
```
mschwoer, line 248: mb give a short comment how `match` looks like in the two cases

alphabase/constants/atom.py
```diff
@@ -200,3 +204,77 @@ def calc_mass_from_formula(formula: str):
...
+    def __add__(self, other):
+        result = ChemicalCompositonFormula("")
+        result.elements = defaultdict(int)
+        for element in set(self.elements.keys()) | set(other.elements.keys()):
+            result.elements[element] = self.elements[element] + other.elements[element]
+        return result
+
+    def __sub__(self, other):
+        result = ChemicalCompositonFormula("")
+        result.elements = defaultdict(int)
```
mschwoer, line 277: the knowledge that `elements` is `defaultdict` is currently at three places .. 
suggestions:
either introduce a small method
`_init_elements(): return defaultdict(int)` and use it at these places.
or change the constructor, e.g.
```
    def __init__(self, formula = None):
        self.elements = defaultdict(int) if formula is None else self._parse_formula(formula)
```
I prefer the second option as it also avoids the initialization with `ChemicalCompositonFormula("")`

alphabase/constants/modification.py
```diff
@@ -376,11 +406,12 @@ def _add_a_new_modification(
...
             "unimod_id",
+            "smiles",
         ],
-    ] = [mod_name, composition, modloss_composition, "User-added", 0]
+    ] = [mod_name, composition, modloss_composition, "User-added", 0, smiles]
     MOD_DF.loc[mod_name, ["mass", "modloss"]] = (
-        calc_mass_from_formula(composition),
-        calc_mass_from_formula(modloss_composition),
+        composition_mass,
+        modloss_mass,
```
mschwoer, line 414: (nit) why introduce variables here? at least, I would move them right above this line ;-)

alphabase/smiles/smiles.py
```diff
@@ -0,0 +1,116 @@
...
+
+from alphabase.constants.aa import AA_Formula
+from alphabase.constants.modification import MOD_DF
+
+N_TERM_PLACEHOLDER = "[Xe]"
+C_TERM_PLACEHOLDER = "[Rn]"
+
+MOD_DF_SMILES = MOD_DF[MOD_DF["smiles"] != ""]
+
+aa_smiles = {
```
mschwoer, line 15: I know that import-level logic is sort of a thing around here :-D
Could we turn all this into a class, and do these operations only when needed?

extra_requirements/development.txt
```diff
@@ -32,3 +32,4 @@ pyahocorasick
 pytest
 pre-commit==3.7.0
 nbmake==1.5.3
+rdkit==2024.3.3
```
mschwoer, line 35: is this needed here? it's already in `requirements.txt`

mschwoer (approved):

alphabase/smiles/smiles.py
GeorgWa: I agree. I think it was already suboptimal for the registry pattern and more so for more involved calculations like `Chem.MolFromSmiles()`

alphabase/smiles/smiles.py
```diff
@@ -0,0 +1,116 @@
...
+    for name, smiles in zip(MOD_DF_SMILES["mod_name"], MOD_DF_SMILES["smiles"])
+    if name not in n_term_modifications and name not in c_term_modifications
+}
+
+
+def validate_correct_args(
+    aa_smiles: str, n_term_mod: Optional[str] = None, c_term_mod: Optional[str] = None
+) -> None:
+    """
+    Validate the amino acid, N-terminal modification, and C-terminal modification.
```
GeorgWa, line 43: Please use Numpy style docstrings in alphaBase :)
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

alphabase/constants/modification.py
```diff
@@ -362,12 +362,42 @@ def calc_modloss_mass(
         return _calc_modloss(mod_losses[::-1])[-3:0:-1]
 
 
+def _check_mass_sanity(
```
GeorgWa, line 365: can you add dosctrings here?

GeorgWa (approved):
I think with numpy style docstrings and without import level actions we are good to merge 👍🏻

jalew188 (approved):
LGTM

alphabase/constants/atom.py
boopthesnoot: In this case `mol` will be `None`, so we can track the failed molecules only outside of this method
Alternatively, I can modify it to take a SMILES string as the input, so we could have control over meaningful logging

mschwoer: `if not ""` would also trigger the log ..

boopthesnoot: Yes, and this is the behaviour that I want to have, I'm just saying that there is no use in using the format string here, unless changing the logic of the argument and going from smiles instead of going from rdkit mol object

GeorgWa:
I just catched that the dtype of unimod column in the modification.tsv changed to float. Can we move back?

alphabase/constants/atom.py
boopthesnoot: Rewrote to take smiles as an argument, so we can have more meaningful error messages

alphabase/smiles/smiles.py
boopthesnoot: Rewrote this into a class

alphabase/constants/atom.py
boopthesnoot: Done!

alphabase/constants/atom.py
```diff
@@ -200,3 +204,195 @@ def calc_mass_from_formula(formula: str):
...
+        ValueError
+            If the SMILES string is invalid and can't be converted to an RDKit molecule.
+        """
+        mol = Chem.MolFromSmiles(smiles)
+        if not mol:
+            raise ValueError(f"Invalid RDKit molecule: {smiles}")
+        formula = rdMolDescriptors.CalcMolFormula(
+            mol, separateIsotopes=True, abbreviateHIsotopes=False
+        )
+        formula = formula.replace("[1H]", "H")
```
mschwoer, line 254: could you add a comment why this is required? is this a common substitution? is it proteomics- or alphabase-specific?

alphabase/smiles/smiles.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        if n_term_mod:
+            n_mod_smiles = self.n_term_modifications[n_term_mod]
+            n_mod_mol = Chem.MolFromSmiles(n_mod_smiles)
+            mol = ReplaceSubstructs(mol, n_term_placeholder_mol, n_mod_mol)[0]
+
+            if "Dimethyl" in n_term_mod:
+                mol = ReplaceSubstructs(mol, n_term_placeholder_mol, n_mod_mol)[0]
+            else:
+                mol = ReplaceSubstructs(
```
mschwoer, line 251: (nit) a bit more concise, but not sure if easier to read

```python
replaceAll = True
if n_term_mod:
    n_mod_smiles = self.n_term_modifications[n_term_mod]
    n_mod_mol = Chem.MolFromSmiles(n_mod_smiles)
    mol = ReplaceSubstructs(mol, n_term_placeholder_mol, n_mod_mol)[0]

    if "Dimethyl" in n_term_mod:
        return ReplaceSubstructs(mol, n_term_placeholder_mol, n_mod_mol)[0]

    replaceAll = False
    
return ReplaceSubstructs(
        mol,
        n_term_placeholder_mol,
        Chem.MolFromSmiles("[H]", sanitize=False),
        replaceAll=replaceAll,
    )[0]
```

alphabase/smiles/smiles.py
```diff
@@ -0,0 +1,293 @@
...
+
+        Returns
+        -------
+        Chem.Mol
+            The modified molecule.
+        """
+        if n_term_mod:
+            n_mod_smiles = self.n_term_modifications[n_term_mod]
+            n_mod_mol = Chem.MolFromSmiles(n_mod_smiles)
+            mol = ReplaceSubstructs(mol, n_term_placeholder_mol, n_mod_mol)[0]
```
mschwoer, line 246: I would rename that first `mol` to `original_mol` or smth

alphabase/constants/aa.py
```diff
@@ -11,12 +11,14 @@
     parse_formula,
     reset_elements,
 )
-from alphabase.yaml_utils import load_yaml
 
 # We use all 128 ASCII code to represent amino acids for flexible extensions in the future.
 # The amino acid masses are stored in 128-lengh array :py:data:`AA_ASCII_MASS`.
 # If an ASCII code is not in `AA_Formula`, the mass will be set as a large value to disable MS search.
-AA_Formula: dict = load_yaml(os.path.join(CONST_FILE_FOLDER, "amino_acid.yaml"))
+AA_Formula: pd.DataFrame = pd.read_csv(
```
mschwoer, line 18: just thinking out loud: given that this is a breaking change anyway, should be take the change and rename it to
`aa_formula`? (or `_aa_formula`) if it's supposed to be private?

alphabase/constants/atom.py
```diff
@@ -200,3 +204,195 @@ def calc_mass_from_formula(formula: str):
...
+        The chemical formula as a string.
+
+    Returns
+    -------
+    None
+    """
+
+    def __init__(self, formula=None):
+        self.elements = (
+            defaultdict(int) if formula is None else self._parse_formula(formula)
```
mschwoer, line 225: I am still not sure if having a `defaultdict` as instance variable type is a good idea (I guess it was mine) .. I think the user of the class will not know what it is, and on wrong use, there will be no `KeyErrors` but much less visible zeros..
what do you think?

As a quick remedy, one could convert the `defaultdict` to `dict` when returning from `parse_formula` & `parse_rdkit_formula`

mschwoer (approved):
LGTM! this PR definitely puts smiles in peoples faces ;-)

alphabase/constants/atom.py
boopthesnoot: If 'abbreviateHIsotopes=True' we get a D instead of [2H] and the modification.tsv expects [2H]. As a side effect we now get [1H] instead of regular hydrogens, so I gotta substitute them back.

alphabase/constants/atom.py
boopthesnoot: added a method to make sure elements of the defaultdict are valid atom names

alphabase/smiles/smiles.py
boopthesnoot: would not work with the current implementation

alphabase/smiles/smiles.py
boopthesnoot: made it even more concise

PR #200, opened 2024-07-17
smiles extra and decorator for rdkit functions

GeorgWa:
This allows to install rdkit only when the smiles extra is defined.
All RDKit dependent functions will give a meaningfull error message if called:
```
RDKit is required for the function '{func.__name__}' 
Please install it with 'pip install alphabase[smiles]'
```


alphabase/peptide/smiles.py
```diff
@@ -0,0 +1,21 @@
...
+            from rdkit import Chem
+        except ImportError:
+            raise ImportError(
+                f"RDKit is required for the function '{func.__name__}'. "
+                "Please install it with 'pip install alphabase[smiles]'"
+            )
+        return func(*args, **kwargs)
+    return wrapper
+
+@requires_rdkit
```
mschwoer, line 19: this is not code that is actually used, right?

alphabase/peptide/smiles.py
```diff
@@ -0,0 +1,21 @@
+import functools
+
+def requires_rdkit(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            from rdkit import Chem
```
mschwoer: is there another option to find out that the package is there, without actually importing it? this would avoid the side effect of this decorator

extra_requirements/smiles.txt
```diff
@@ -0,0 +1 @@
+rdkit>=2024.3.3
```
mschwoer: fix version? :-)

alphabase/peptide/smiles.py
mschwoer: also, this will only work if the imports are done within the functions not for top-level imports, right?

GeorgWa: You could use `importlib.util.find_spec("rdkit") is not None` which still invokes the import machinery to some degree.

On the second point, I think there is no way around this.
If you have a module level import you will always fail during initialization. 
Maybe it's possible with some heavy import magic but I'm not aware.

alphabase/peptide/smiles.py
GeorgWa: yes, just as a reference for @boopthesnoot

alphabase/peptide/smiles.py
mschwoer: okay, but can we then not just catch the import error in the `smiles` module, and change it to the user-friendly error message?

jalew188: Yes, I prefer to catch ImportError once this module is imported

jalew188:
What is the connection between this PR and #199

jalew188:
If smiles is imported in many other modules, i think it is better to place rdkit in the main requirement

jalew188 (approved):

Issue #198, opened 2024-07-16
Save precursor_df into HDF5 as a single dataframe instead of two


jalew188:
Currently, for faster access in the speclib HDF, numeric columns are saved into `precursor_df` and string columns are saved into `mod_seq_df`, which is not easy to read. It is better to use a single dataframe instead of two.

PR #201, opened 2024-07-17
#198 a new param save_mod_seq_in_other_df=False in SpecLibBase.save_hdf

jalew188:


alphabase/spectral_library/base.py
```diff
@@ -581,51 +582,71 @@ def load_df_from_hdf(self, hdf_file: str, df_name: str) -> pd.DataFrame:
         """
         return self._get_hdf_to_load(hdf_file).__getattribute__(df_name).values
 
-    def save_hdf(self, hdf_file: str):
+    def save_hdf(self, hdf_file: str, save_mod_seq_in_other_df: bool = False):
```
GeorgWa, line 585: I think this is good. This shifts the responsibility to the writer of HDFs.
By default we have the simple case of 1) single df 2) loading modseq for old style dfs.

GeorgWa (approved):
LGTM

alphabase/spectral_library/base.py
jalew188: Yes

PR #203, opened 2024-07-17
update unimod to tsv

GeorgWa:
I was looking for unimod 2062 which was still missing from the modifications.tsv

I've updated the notebook so it:
- automatically converts whitespaces to underscore
- adds all new modifications not yet in modifications.tsv
- adds a `smiles` column


jalew188 (approved):
Keep in mind that, to add user-defined modifications, we also allow users to add smiles as an option.

PR #213, opened 2024-08-05
Development

jalew188:
Release of 1.3.0


jalew188:
> Hi, what do you think about raising the version to 2.0 with the new modifications?

We have discussed about this before. Before alphabase's paper publish, we can still use 1.x.x for breaking changes. This can prevent version 2000.x.x due to too many breaking changes as we still modify alphabase time to time.

jalew188:
As long as the paper is published (BioRxiv), we will release the version 2.0.0 and then we will follow the right version control criteria.

GeorgWa:
Okay 👍

GeorgWa (approved):
Lgtm

Issue #215, opened 2024-08-08
Ion mobility column of spectronaut report not being read in by psm_reader 


mlorenz49:
When trying to read in the output of spectronaut_report file, the column called "FG.ApexIonMobility" is not being imported with the rest of the data (the column seems to be dropped). Trying out the following seems to read in the information into self._psm_df.

```
from alphabase.psm_reader.dia_psm_reader import psm_reader_yaml
psm_reader_yaml["spectronaut_report"]["column_mapping"]["mobility"]="FG.ApexIonMobility"
```

However, the data type of the entries in the "mobility" column are strings and not floats. This leads to a downstream error in the `alphabase.peptide.mobility.mobility_to_ccs_bruker` function, as here the strings are multiplied with a float value (`CCS_IM_COEF`). Ive attached a screenshot with the issue.

Is there a way to import the column correctly?

**Screenshots**
![image](https://github.com/user-attachments/assets/bc149125-2d8c-4216-9629-ffa4f7b585df)

**Version:**
 - Installation Type pip install
   - Platform information
     - system windows
   - Python information:
     - alphabase version 1.2.5
     - [other packages]

jalew188:
@mlorenz49 Please checkout #216 for testing

PR #216, opened 2024-08-09
#215 add FG.ApexIonMobility for SPN report reader

jalew188:


mlorenz49:
That seems to be correct, I don't have the rights to merge the pull request though.

mlorenz49 (approved):

jalew188:
@mlorenz49 does it work?

mlorenz49:
Sadly not, the same error as described in the issue appears when running. I attached a screenshot where we can see that the mobility column is being read in at first and is available in self._psm_df. The error is thrown when reaching the function mobility.mobility_to_ccs_for_df in line 509 of the psm_reader.py file.

After running the command `self._psm_df["mobility"][0]` in the debugger the string value `'0.97584784'` is returned.

Changing the value in psm_reader.yaml from `'mobility': ['FG.ApexIonMobility']` to `'mobility': 'FG.ApexIonMobility'` doesnt solve the issue either.

image

mlorenz49:
I found the error:

code on line 176 `df = pd.read_csv(filename, sep=self.csv_sep`) was causing an issue while trying to read in rows with NaN values in the newly added mobiliy column of the spectronaut_reader type. Removing the option `keep_default_na=False` correctly imports the column as float and not as string. NA values are set to 0.0 (not sure how good of a solution that is)

I ran all the tests and they passed. I dont have the rights to push to this branch though.

PR #220, opened 2024-09-23
Peptides SMILES representation

boopthesnoot:
Added functionality to get a SMILES string representing a peptide, optionally with PTMs.
See jupyter notebook for the example usage :)


alphabase/smiles/peptide.py
```diff
@@ -0,0 +1,196 @@
...
+
+class PeptideSmilesEncoder:
+    """
+    A class to encode peptide sequences into SMILES strings or RDKit molecule objects.
+    """
+
+    def __init__(self):
+        self.amino_acid_modifier = AminoAcidModifier()
+
+    def encode_peptide(
```
GeorgWa, line 16: Should we have a function `peptide_to_smiles` and `peptide_to_mol` to return the rdkit object if wanted?

alphabase/smiles/peptide.py
```diff
@@ -0,0 +1,196 @@
...
+            n_term_mod=n_term_mod,
+        )
+        for idx in range(1, len(amino_acid_mols)):
+            peptide_mol = self._connect_amino_acids(peptide_mol, amino_acid_mols[idx])
+            peptide_mol = self.amino_acid_modifier._apply_n_terminal_modification(
+                peptide_mol,
+                n_term_placeholder_mol=n_term_placeholder_mol,
+                n_term_mod=None,
+            )
+        peptide_mol = self.amino_acid_modifier._apply_c_terminal_modification(
```
GeorgWa, line 114: just for understanding, when are the N-term and C-term placeholders removed?
Or is this solved by concatenating them and applying a C-term water, N-Term ammonia modification?

GeorgWa (approved):
Looks good to me

alphabase/smiles/peptide.py
boopthesnoot: The N-term modification is applied to the first amino acid (what is line 102 now), after that, we start growing the peptide, and after each addition of the next amino acid we use `_apply_n_terminal_modification` to get rid of the second N-term placeholder (line 109). When we add the last AA, we modify the C-term of the resulting peptide (line 114).

jalew188: Do smiles start with N-terminal amino group of peptides?

alphabase/smiles/peptide.py
boopthesnoot: refactored to have both those methods

alphabase/smiles/peptide.py
boopthesnoot: The string itself? Not necessarily, but we shouldn't work with it with string methods anyway. Please check out the structures in the jupyter notebook, I hope I didn't mess something up with the order or something

jalew188: Yes, for graph, it is not necessary.

jalew188 (approved):

alphabase/smiles/peptide.py
```diff
@@ -0,0 +1,225 @@
...
+        c_term_mod = None
+
+        if 0 in mods:
+            n_term_mod = mods[0]
+        if -1 in mods:
+            c_term_mod = mods[-1]
+        # Process each amino acid in the sequence
+        for idx, aa in enumerate(sequence):
+            if idx + 1 in mods:
+                aa_smiles = self.amino_acid_modifier.ptm_dict.get(mods[idx + 1], None)
```
jalew188, line 114: How many PTMs are supported nwo?

jalew188: Looks great!!

PR #225, opened 2024-10-30
open hdf file "r" mode on read operations

mschwoer:
Open HDF file in "r" mode if only read operations are done. 

This is to reduce errors ` ERROR: [Errno 11] Unable to synchronously open file (unable to lock file, errno = 11, error message = 'Resource temporarily unavailable')` on HPC systems when several instances try to access the same file.

@jalew188 alternatively one could substitute the `mode=a` with `mode="r" if self._read_only else "a"` if you see any implications here

### DIFF
--- a/alphabase/io/hdf.py
+++ b/alphabase/io/hdf.py
@@ -45,7 +45,7 @@ class HDF_Object:
 
     @property
     def metadata(self):
-        with h5py.File(self.file_name, "a") as hdf_file:
+        with h5py.File(self.file_name) as hdf_file:
             return dict(hdf_file[self.name].attrs)
 
     def __init__(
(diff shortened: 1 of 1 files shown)


jalew188 (approved):
This is better in terms of multi-access.

jalew188:
> @jalew188 alternatively one could substitute the `mode=a` with `mode="r" if self._read_only else "a"` if you see any implications here

For me, this is fine as 'a' is not easy to handle in our senario.

alphabase/io/hdf.py
```diff
@@ -45,7 +45,7 @@ def editing_mode(self, read_only: bool = False, truncate: bool = True):
 
     @property
     def metadata(self):
-        with h5py.File(self.file_name, "a") as hdf_file:
+        with h5py.File(self.file_name) as hdf_file:
```
GeorgWa, line 48: Do I understand it correctly that this is the same as`h5py.File(self.file_name, 'r')` ?
Can we add the `'r'` explicitly so it's clear that this was a deliberate choice.

GeorgWa (approved):
LGTM, should solve some issues we had

alphabase/io/hdf.py
mschwoer: hm, isn't not adding "a" deliberate enough? ;-)

PR #227, opened 2024-10-31
Adapt to new release workflow

mschwoer:
remove release workflow as alphabase does not need to be installed


GeorgWa (approved):
LGTM, altough we are not "removing the release workflow" are we? ;)

PR #228, opened 2024-11-04
updated psm-reader settings for maxquant

GeorgWa:
Extend psm-reader yaml for MQ reader


alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -38,6 +38,7 @@ maxquant:
     'scan_num':
       - 'Scan number'
       - 'MS/MS scan number'
+      - 'MS/MS Scan Number'
```
mschwoer, line 41: how does the parser cope with those multiple values? takes the first that it finds?

maybe you can add a comment, from which MQ version on this is valid,
e.g.
`- 'MS/MS Scan Number' # MQ version >=1.2.3`

mschwoer (approved):

PR #229, opened 2024-11-04
remove maxquant MBR rows

GeorgWa:
currently alphabase fails for maxquant output which contains MBR PSMs.
This removes MBR PSMs and prints an explicit warning

### DIFF
--- a/alphabase/psm_reader/maxquant_reader.py
+++ b/alphabase/psm_reader/maxquant_reader.py
@@ -12,6 +12,7 @@ from alphabase.psm_reader.psm_reader import (
     psm_reader_yaml,
 )
 
+# make sure all warnings are shown
 warnings.filterwarnings("always")
 
 mod_to_unimod_dict = {}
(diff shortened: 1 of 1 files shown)


alphabase/psm_reader/maxquant_reader.py
```diff
@@ -11,6 +12,8 @@
     psm_reader_yaml,
 )
 
+warnings.filterwarnings("always")
```
mschwoer, line 15: could the scope of this be reduced, e.g. to a specific code block and/or specific warnings?

alphabase/psm_reader/maxquant_reader.py
```diff
@@ -245,6 +248,20 @@ def _load_file(self, filename):
...
+        if "scan_num" in mapped_columns:
+            scan_num_col = mapped_columns["scan_num"]
+            no_ms2_mask = df[scan_num_col] == ""
+            if np.sum(no_ms2_mask) > 0:
+                warnings.warn(
+                    f"Maxquant psm file contains {np.sum(no_ms2_mask)} MBR PSMs without MS2 scan. This is not yet supported and rows containing MBR PSMs will be removed."
+                )
+                df = df[~no_ms2_mask]
+                df.reset_index(drop=True, inplace=True)
+                df[scan_num_col] = df[scan_num_col].astype(int)
```
mschwoer, line 263: if we don't hit the `if "scan_num" in mapped_columns:` block: will the type of `scan_num_col` still be `int`?
or turning it around: we should make sure all columns' dtypes are consistent, and independent of the inner logic of the reader.

alphabase/psm_reader/maxquant_reader.py
```diff
@@ -245,6 +248,20 @@ def _load_file(self, filename):
         self._find_mod_seq_column(df)
         df = df[~pd.isna(df["Retention time"])]
         df.fillna("", inplace=True)
+
+        # remove MBR PSMs as they are currently not supported and will crash import
+        mapped_columns = self._find_mapped_columns(df)
+        if "scan_num" in mapped_columns:
+            scan_num_col = mapped_columns["scan_num"]
+            no_ms2_mask = df[scan_num_col] == ""
+            if np.sum(no_ms2_mask) > 0:
```
mschwoer, line 257: (nit)
`if (num_no_ms2_mask := np.sum(no_ms2_mask)) > 0:`

mschwoer (approved):

alphabase/psm_reader/maxquant_reader.py
GeorgWa: Maybe you know more about this.
This code is needed to enforce that warning are shown because it's somehow not happening.
So it's not removing warnings but showing them.

mschwoer: ah, I misinterprested this as "filter always" .. then it's fine, maybe add the information that warnings are otherwise not shown as a comment ..

PR #232, opened 2024-11-13
Refactor readers i

mschwoer:
Some refactorings of the PSM readers

- use `_psm_df` iternally instead of `psm_df` (first commit)
- simplify column_mapping initialization (second commit)

- Introduce string constants to access the data frames in the reader classes.
Benefits: the relation between the different locations where columns are accessed becomes transparent.


alphabase/psm_reader/keys.py
```diff
@@ -0,0 +1,71 @@
...
+    def get_values(cls):
+        """Get all user-defined string values of the class."""
+        return [
+            value
+            for key, value in cls.__dict__.items()
+            if not key.startswith("__") and isinstance(value, str)
+        ]
+
+
+class PsmDfCols(metaclass=ConstantsClass):
```
jalew188, line 16: Shall we make this const like`PSM_DF_COLS`?

jalew188:
LGTM!

lucas-diedrich:
scVI tools (really solid library, part of the scVerse - HA!) inherits from the named tuple class

- https://github.com/scverse/scvi-tools/blob/998ea3dabf5434dcacaf78bb707ef0ca54f10593/src/scvi/data/_constants.py#L4
- https://github.com/scverse/scvi-tools/blob/998ea3dabf5434dcacaf78bb707ef0ca54f10593/src/scvi/_constants.py#L2

GeorgWa (approved):
Looks good!
Two comments from my side:
- although the readers have the psm_df the same data frame is called precursor_df in the rest of the codebase when present in spectral libraries etc. we could unify this at some point. 
- some of the columns used here are temporary or redundant as they will be removed later. That’s not great software engineering wise but maybe we can mark them with an underscore.

alphabase/psm_reader/keys.py
GeorgWa: I think a class is not a bad idea. In alphadia I want to have a superset of this. So I could inherit and extend it.

alphabase/psm_reader/keys.py
```diff
@@ -0,0 +1,71 @@
...
+    PEPTIDE_FDR = "peptide_fdr"
+    PROTEIN_FDR = "protein_fdr"
+
+    RAW_NAME = "raw_name"
+    CHARGE = "charge"
+    PROTEINS = "proteins"
+
+    SCAN_NUM = "scan_num"
+    PRECURSOR_MZ = "precursor_mz"
+    DIANN_SPEC_INDEX = "diann_spec_idx"
```
GeorgWa, line 48: Should we get rid of temp/reader specific columns and rather process them in place etc.?

GeorgWa (approved):
LGTM

alphabase/psm_reader/keys.py
lucas-diedrich: I think the class is a good idea, looks similar to the way scVerse packages handle this: 

- https://github.com/scverse/scvi-tools/blob/54ba452f50db0b6a3ca8f4e46837e981362e01d4/src/scvi/_constants.py#L3

lucas-diedrich (approved):

alphabase/psm_reader/keys.py
GeorgWa: Absolutely! I though more about columns which are not returned in the final object but just for 'communication' within a processing step

alphabase/psm_reader/keys.py
mschwoer: if it's only about the naming: I'd prefer to stick to the current name, first not to violoate the convention, second because `PSM_DF_COLS.PRECURSOR_MZ` is imho harder to read than `PsmDfCols.PRECURSOR_MZ` (personal preference)

mschwoer:
@GeorgWa thanks for the heads up regarding precursor_df & the unused columns, noted.

alphabase/psm_reader/keys.py
jalew188: OK!

PR #233, opened 2024-11-14
Linting i

mschwoer:
add a second linting with a stricter ruleset, apply autofixes

### DIFF
--- a/ruff-lint-psm-readers.toml
+++ b/ruff-lint-psm-readers.toml
@@ -4,15 +4,16 @@ select = [
     "ALL"
 ]
 
-# TODO this is not great but it is a workaround for now:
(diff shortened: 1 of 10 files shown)


ruff-lint-psm-readers.toml
```diff
@@ -0,0 +1,65 @@
+# specific ruff configuraton for the psm readers (which have stricter rules)
+[lint]
+select = [
+    "ALL"
+]
+
+# TODO this is not great but it is a workaround for now:
+exclude = [
+    "**/psm_reader/*.py",
+    "**/spectral_library/reader.py",
```
mschwoer, line 10: these two exclusions will go away after all rules are satisfied (that's the point of this excersise ;-))

jalew188 (approved):
LGTM

PR #235, opened 2024-11-15
Linting iii

mschwoer:
manual fixes for a first bunch of linting rules ..
see individual commits

edit: made all start imports in notebooks explicit as tests started to fail due to removal of `os` from some readers


alphabase/psm_reader/alphapept_reader.py
```diff
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
```
GeorgWa, line 1: Is this a thing? that `Path` is prefered over os?

alphabase/psm_reader/maxquant_reader.py
```diff
@@ -18,8 +18,8 @@
 warnings.filterwarnings("always")
 
 mod_to_unimod_dict = {}
-for mod_name, unimod_id in MOD_DF[["mod_name", "unimod_id"]].values:
-    unimod_id = int(unimod_id)
+for mod_name, unimod_id_ in MOD_DF[["mod_name", "unimod_id"]].to_numpy():
```
GeorgWa, line 21: is this suffix thing a known pattern? I only now it as private prefix

alphabase/psm_reader/sage_reader.py
```diff
@@ -292,7 +292,7 @@ def _lookup_modification(
         The name of the matched modification in alphabase format.
 
     """
-    mass_distance = mod_annotated_df["mass"].values - mass_observed
+    mass_distance = mod_annotated_df["mass"].to_numpy() - mass_observed
```
GeorgWa, line 295: This is a bit of a tricky change. It might have some performance implications as it performas a copy.

alphabase/spectral_library/reader.py
```diff
@@ -169,62 +168,62 @@ def _get_fragment_intensity(self, lib_df: pd.DataFrame):
...
                     LibPsmDfCols.FRAGMENT_LOSS_TYPE,
                     LibPsmDfCols.FRAGMENT_CHARGE,
                     LibPsmDfCols.FRAGMENT_INTENSITY,
                 ]
-            ].values:
-                if frag_type in "abc":
-                    frag_num -= 1
-                elif frag_type in "xyz":
-                    frag_num = nAA - frag_num - 1
+            ].to_numpy():
```
GeorgWa, line 185: same as before

GeorgWa (approved):
LGTM
I think the `Path() / Path()` syntax is quite ugly and missuses python magic methods :D but if the community thinks this is the way to go I'm happy to do so.
I'm also not a big fan of ther new read only `.to_numpy()` buit I guess I have to live with this.

alphabase/psm_reader/alphapept_reader.py
mschwoer: yes.. imho it makes path operations much more readable, 
I especially like the overloading of the `/` operator when it comes to using it (e.g. `Path("/home/") / "user"`).
I especially dislike the overloading of the `/` operator when it comes to mocking it :-)

alphabase/psm_reader/maxquant_reader.py
mschwoer: I use it to avoid name clashes, e.g. with built-ins or other variables

alphabase/psm_reader/sage_reader.py
mschwoer: If you wish, we can disable the rule and revert? however, pandas docs recommend using `to_numpy()` 
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.values.html

PR #242, opened 2024-11-18
Refactor readers iii

mschwoer:
simplify the modification mapping handling
(see individual commits for details)

### DIFF
--- a/alphabase/psm_reader/psm_reader.py
+++ b/alphabase/psm_reader/psm_reader.py
@@ -259,7 +259,7 @@ class PSMReaderBase(ABC):
             self.modification_mapping = copy.deepcopy(modification_mapping)
 
         self._mods_as_lists()
-        self._reverse_mod_mapping()
+        self.rev_mod_mapping = self._get_reversed_mod_mapping()
 
     def _init_modification_mapping(self) -> None:
         self.modification_mapping = {}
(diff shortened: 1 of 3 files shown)


alphabase/psm_reader/psm_reader.py
```diff
@@ -219,30 +221,36 @@ def add_modification_mapping(self, modification_mapping: dict) -> None:
...
+        ----------
+        modification_mapping:
+            If dictionary: the current modification_mapping will be overwritten by this.
+            If str: the parameter will be interpreted as a reader type, and the modification_mapping is read from the
+                "modification_mapping" section of the psm_reader_yaml
+
+        """
         if modification_mapping is None:
             self._init_modification_mapping()
         elif isinstance(modification_mapping, str):
```
mschwoer, line 256: @jalew188 here, a lot of hidden complexity arises (also connected to `_init_modification_mapping`) as the "column_mapping" parameter in the yaml is overloaded (could either be str or dict).
Could we introduce a (mutually exclusive ) "column_mapping_type` in the yaml, that served to reference another reader?

alphabase/psm_reader/psm_reader.py
```diff
@@ -257,33 +265,39 @@ def set_modification_mapping(
         else:
             self.modification_mapping = copy.deepcopy(modification_mapping)
 
-        self._mods_as_lists()
-        self._reverse_mod_mapping()
+        self._str_mods_to_lists()
```
lucas-diedrich, line 268: Is there a reason why this attribute is mutated in place, while the others are reassigned?
(see e.g. line 269)

alphabase/psm_reader/psm_reader.py
jalew188: Yes, we shall then add a new method: `add_new_column_mapping_item` to handle user-defined columns

jalew188: Both modification_mapping and column_mapping is not well designed in alphabase readers. We need a better way.

alphabase/psm_reader/psm_reader.py
jalew188: I guess for readability? Complex Inplace operations are hard to read and track

mschwoer: good catch @lucas-diedrich , I was wondering too .. it's as @jalew188 suspected indeed the limited complexity that made me keep `str_mods_to_lists` as in-place .. but more based on a gut feeling rather than on a strict reasoning .. I could change it for consistency if you like?

alphabase/psm_reader/psm_reader.py
```diff
@@ -219,30 +221,36 @@ def add_modification_mapping(self, modification_mapping: dict) -> None:
...
-                    self.modification_mapping[key].append(val)
-                else:
-                    self.modification_mapping[key].extend(val)
-            elif isinstance(val, str):
-                self.modification_mapping[key] = [val]
+            if isinstance(val, str):
+                new_modification_mapping[key].append(val)
             else:
-                self.modification_mapping[key] = val
+                new_modification_mapping[key].extend(val)
```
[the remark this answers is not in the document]
mschwoer: Note that `new_modification_mapping` is `defaultdict(list)`, so we don't need the `if key in ..`  part .. was this the change you were referring to?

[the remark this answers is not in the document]
jalew188: I see, that would be fine now

alphabase/psm_reader/psm_reader.py
lucas-diedrich: I see, thank you for the explanation 👍 So I guess the alternative would then be something like 
```python
self.modification_mapping = _str_mods_to_lists(self.modification_mapping)
```

With
```python
def _str_mods_to_lists(self, modification_mapping: dict) -> dict[str, list[str]]:
     """Convert all single strings to lists containing one item in self.modification_mapping."""
     for mod, val in list(modification_mapping.items()):
         if isinstance(val, str):
             modification_mapping[mod] = [val]
     return modification_mapping
```

I don't have a strong opinion on this, and the current way looks great, so no need to change it!

alphabase/psm_reader/psm_reader.py
GeorgWa: I absolutely agree. I think the modification mapping is already quite close but we need to make sure it works the same where everywhere. It's also not yet compatible with custom mods and MP.

GeorgWa (approved):
LGTM

jalew188 (approved):
This part looks good to me

PR #245, opened 2024-11-19
fix identation in MAxQuantReader._load_file

mschwoer:
fixes bug introduced here: https://github.com/MannLabs/alphabase/pull/229


GeorgWa (commented):
LGTM

PR #246, opened 2024-11-19
Refactor readers iv

mschwoer:
Trying to simplify the inheritance structure
- make `self.csv_sep` a ordinary variable
- move some logic up to the reader base class
- extract static helper methods to utils.py, add tests

### DIFF
--- a/alphabase/psm_reader/psm_reader.py
+++ b/alphabase/psm_reader/psm_reader.py
@@ -11,84 +11,17 @@ import numpy as np
 import pandas as pd
 
 from alphabase.constants._const import CONST_FILE_FOLDER
-from alphabase.constants.modification import MOD_DF
 from alphabase.peptide import mobility
 from alphabase.peptide.precursor import reset_precursor_df, update_precursor_mz
 from alphabase.psm_reader.keys import PsmDfCols
+from alphabase.psm_reader.utils import (
+    MOD_TO_UNIMOD_DICT,
+    get_mod_set,
+    keep_modifications,
+    translate_other_modification,
+)
 from alphabase.utils import _get_delimiter
 from alphabase.yaml_utils import load_yaml
-
-mod_to_unimod_dict = {}
-for mod_name, unimod_id_ in MOD_DF[["mod_name", "unimod_id"]].to_numpy():
-    unimod_id = int(unimod_id_)
-    if unimod_id in (-1, "-1"):
-        continue
-    if mod_name[-2] == "@":
-        mod_to_unimod_dict[mod_name] = f"{mod_name[-1]}(UniMod:{unimod_id})"
-    else:
-        mod_to_unimod_dict[mod_name] = f"_(UniMod:{unimod_id})"
-
-
-def translate_other_modification(mod_str: str, mod_dict: dict) -> str:
-    """Translate modifications of `mod_str` to the AlphaBase format mapped by mod_dict.
-
-    Parameters
-    ----------
-    mod_str : str
-        mod list in str format, seperated by ';',
-        e.g. ModA;ModB
-    mod_dict : dict
-        translate mod dict from others to AlphaBase,
-        e.g. for pFind, key=['Phospho[S]','Oxidation[M]'],
-        value=['Phospho@S','Oxidation@M']
-
-    Returns
-    -------
-    str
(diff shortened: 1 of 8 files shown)


tests/unit/psm_reader/test_utils.py
```diff
@@ -0,0 +1,115 @@
...
+        "_[Phospho]",
+    ]
+
+
+def test_get_extended_modifications_with_modifications_starting_with_underscore_square_bracket():
+    assert get_extended_modifications(["_[Phospho]"]) == [
+        "[Phospho]",
+        "_(Phospho)",
+        "_[Phospho]",
+    ]
```
mschwoer, line 115: @jalew188 just by symmetry, I would expect `(Phospho)` here too?

alphabase/psm_reader/utils.py
```diff
@@ -0,0 +1,113 @@
+"""Utility functions for PSM readers."""
```
mschwoer, line 1: only moved code

tests/unit/psm_reader/test_utils.py
jalew188: It should be, but I am not sure what is this testing for.

mschwoer: changed the test data to be more realistic here: https://github.com/MannLabs/alphabase/pull/247/files#diff-49c8564dc5623dcc26f0b191744d709f762671f9d6cad81442985f3f75cb23fa

alphabase/psm_reader/psm_reader.py
```diff
@@ -266,17 +213,41 @@ def set_modification_mapping(
...
+            self._extend_mod_brackets()
+
         self.rev_mod_mapping = self._get_reversed_mod_mapping()
 
     def _init_modification_mapping(self) -> None:
         self.modification_mapping = {}
 
+    def _add_all_unimod(self) -> None:
+        for mod_name, unimod in MOD_TO_UNIMOD_DICT.items():
+            if mod_name in self.modification_mapping:
```
GeorgWa, line 228: could we reuse the logic used to add custom mods here? I feel like adding unimod, adding reader specific mods and adding custom mods is very similar.

alphabase/psm_reader/psm_reader.py
```diff
@@ -266,17 +213,41 @@ def set_modification_mapping(
...
+                self.modification_mapping[mod_name].append(unimod)
+            else:
+                self.modification_mapping[mod_name] = [unimod]
+
+    def _extend_mod_brackets(self) -> None:
```
GeorgWa, line 233: This code (ab in general, not your PR) feels to me like we should start a ModificationMapping class 😆

alphabase/psm_reader/psm_reader.py
```diff
@@ -457,20 +428,9 @@ def _translate_columns(self, origin_df: pd.DataFrame) -> None:
             self._psm_df[PsmDfCols.SPEC_IDX] = self._psm_df[PsmDfCols.SCAN_NUM] - 1
 
     def _transform_table(self) -> None:  # noqa: B027 empty method in an abstract base class
-        """Transform the dataframe format if needed.
+        """Transform the dataframe format if needed, ddd information inplace into self._psm_df.
```
GeorgWa, line 431: (nit) typo: add

GeorgWa (approved):
LGTM

alphabase/psm_reader/psm_reader.py
jalew188: Agree. It also depends on how heavy we are going to use these reader classes. Before, I just used them to get unified training data.

tests/unit/psm_reader/test_utils.py
jalew188: You are right...

jalew188: Maybe currently, this is not used by any readers, but we should add () case

jalew188 (approved):
LGTM

alphabase/psm_reader/psm_reader.py
mschwoer: done here https://github.com/MannLabs/alphabase/pull/247

PR #249, opened 2024-11-20
Refactor readers vii

mschwoer:
separate the file loading and preprocessing

### DIFF
--- a/alphabase/psm_reader/dia_psm_reader.py
+++ b/alphabase/psm_reader/dia_psm_reader.py
@@ -88,7 +88,7 @@ class SwathReader(SpectronautReader):
         )
 
 
-class DiannReader(SpectronautReader):
+class DiannReader(MaxQuantReader):
     """Reader for DIANN data."""
 
     _reader_type = "diann"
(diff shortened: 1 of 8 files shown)


alphabase/psm_reader/dia_psm_reader.py
```diff
@@ -93,7 +88,7 @@ def __init__(  # noqa: PLR0913 many arguments in function definition
         )
 
 
-class DiannReader(SpectronautReader):
+class DiannReader(MaxQuantReader):
```
GeorgWa, line 91: Sorry if I get confused here :D 
Didn't we move to a new name for the MaxQuantReader as base tsv reader?

alphabase/psm_reader/sage_reader.py
```diff
@@ -562,7 +563,7 @@ def _sage_spec_idx_from_scan_nr(scan_indicator_str: str) -> int:
     return int(re.search(r"scan=(\d+)", scan_indicator_str).group(1)) - 1
 
 
-class SageReaderBase(PSMReaderBase):
+class SageReaderBase(PSMReaderBase, ABC):
```
GeorgWa, line 566: why do we need to inherit the ABC here?

GeorgWa (approved):
LGTM

alphabase/psm_reader/sage_reader.py
mschwoer: bc `SageReaderBase` should not be instantiated

PR #250, opened 2024-11-20
Refactor readers viii

mschwoer:
remove some obsolete constructors

### DIFF
--- a/alphabase/psm_reader/dia_psm_reader.py
+++ b/alphabase/psm_reader/dia_psm_reader.py
@@ -1,7 +1,5 @@
 """Readers for Spectronaut's output library and reports, Swath data and DIANN data."""
 
-from typing import List, Optional
-
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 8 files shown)


alphabase/psm_reader/dia_psm_reader.py
```diff
@@ -19,32 +17,8 @@ class SpectronautReader(MaxQuantReader):
...
-            keep_decoy=keep_decoy,
-            mod_seq_columns=mod_seq_columns,
-            fixed_C57=fixed_C57,
-            rt_unit=rt_unit,
-            **kwargs,
-        )
-
-        self._min_max_rt_norm = True
+    _min_max_rt_norm = True
+    _fixed_c57_default = False
```
GeorgWa, line 21: Can we enforce that all of them must be set via an ABC?

alphabase/psm_reader/dia_psm_reader.py
```diff
@@ -19,32 +17,8 @@ class SpectronautReader(MaxQuantReader):
...
-            fdr=fdr,
-            keep_decoy=keep_decoy,
-            mod_seq_columns=mod_seq_columns,
-            fixed_C57=fixed_C57,
-            rt_unit=rt_unit,
-            **kwargs,
-        )
-
-        self._min_max_rt_norm = True
+    _min_max_rt_norm = True
```
GeorgWa, line 20: I think this is something we might rather want to do on an alphabase level as post processing

GeorgWa (approved):
LGTM

alphabase/psm_reader/sage_reader.py
```diff
@@ -633,21 +652,13 @@ def _translate_modifications(self) -> None:
 class SageReaderTSV(SageReaderBase):
     """Reader for Sage output files in TSV format."""
 
-    def __init__(self, *args, **kwargs):
-        """Initialize SageReaderTSV."""
-        super().__init__(*args, **kwargs)
```
jalew188, line 638: Why we don't need base.init()

jalew188 (approved):
LGTM

alphabase/psm_reader/dia_psm_reader.py
mschwoer: the base class defines defaults for them which are overwritten here

PR #252, opened 2024-11-20
Refactor readers x

mschwoer:
move more responsibility to psm_reader.yaml

also, changed the structure of `psm_reader.yaml`: all modification_mappings are now gathered under one key and referenced by readers.. let me know what you think!


alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -262,3 +265,4 @@ sage:
     'peptide_fdr': 'peptide_q'
     'protein_fdr': 'protein_q'
     'decoy': 'is_decoy'
+  modification_mapping_type: 'maxquant'
```
GeorgWa, line 268: I'm not sure if this has an effect. sage uses a different mapping method to my knowledge.

alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -13,13 +13,8 @@ alphapept:
     'raw_name': 'raw_name' #parse from `ms_data.hdf`` file
     'fdr': 'q_value'
     'decoy': 'decoy'
-  modification_mapping:
```
jalew188, line 16: I think we shall show here that if search engines use unimod_id

alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -119,8 +124,7 @@ pfind:
     'uniprot_ids': 'Proteins'
     'fdr': 'Q-value'
     'decoy': ['Target/Decoy', 'Targe/Decoy']
-  modification_mapping:
-    '': ''
+  modification_mapping_type: 'maxquant'
```
jalew188, line 127: pFind does not use MaxQuant-like PTM format? Did this pass the pFind tests?

jalew188 (changes requested):
pFind-specific testing result is required

alphabase/constants/const_files/psm_reader.yaml
mschwoer: good catch .. this works because in pFindReader, _translate_modifications() is overwritten .. will change the yaml

alphabase/psm_reader/maxquant_reader.py
```diff
@@ -138,7 +137,7 @@ def __init__(  # noqa: PLR0913, D417 # too many arguments in function definition
         mod_seq_columns: Optional[List[str]] = None,
         fdr: float = 0.01,
         keep_decoy: bool = False,
-        rt_unit: str = "minute",
```
jalew188, line 141: Removing this means users can only know default rt_unit of search engines from yaml settings, instead of the code or docstrings.

mschwoer: yes, but this has not changed, as also before the value in the `yaml` file could overwrite the value defined here .. so just looking at the code could even give you a wrong impression.

users can read off the rt unit from `self._rt_unit` (which I could make public for that sake)

jalew188 (approved):

PR #253, opened 2024-11-22
introduce ModifiedSequenceReader 

mschwoer:
- simplifying the inheritance structure and avoid overwriting superclass methods with `pass`
- get rid of obsolete checks on empty `origin_df`
- get rid of `_transform_table`
- drop hardcoding of "EG.PrecursorId"
- move `test_data_downloader.py` to `tools/data_downloader.py`


GeorgWa:
Should we maybe change the naming to modified sequence reader?

GeorgWa (approved):
I tried to understand it as good as possible. And it mostly looked like cleanup/smaller refactoring. LGTM

mschwoer:
see https://github.com/MannLabs/alphabase/pull/256

PR #256, opened 2024-11-26
Refactor readers xi

mschwoer:
somehow github got stuck on the original PR: https://github.com/MannLabs/alphabase/pull/253

- simplifying the inheritance structure and avoid overwriting superclass methods with `pass`
- get rid of obsolete checks on empty `origin_df`
- get rid of `_transform_table`
- drop hardcoding of "EG.PrecursorId"
- move `test_data_downloader.py` to `tools/data_downloader.py`


jalew188 (approved):
This part looks good to me

PR #260, opened 2024-12-12
Fix bug tempmmap

sophiamaedler:
fix a bug where passed absolute file checks are not handled correctly


mschwoer (approved):
Thanks for fixing this!

PR #261, opened 2024-12-17
Refactor tempmmap

mschwoer:
- put tempmmap.py under test (done before the refactoring)
- create temporary directories not on module import but on demand
- avoid unnecessary creation of temporary directory at program exit

### DIFF
--- a/alphabase/io/tempmmap.py
+++ b/alphabase/io/tempmmap.py
@@ -85,14 +85,16 @@ def _get_file_location(abs_file_path: str, overwrite=False) -> str:
 
     # ensure that the filename conforms to the naming convention
     if not os.path.basename(abs_file_path).endswith(".hdf"):
-        raise ValueError("The chosen file name needs to end with .hdf")
+        raise ValueError(
+            f"The chosen file name '{os.path.basename(abs_file_path)}' needs to end with .hdf"
+        )
 
     # ensure that the directory in which the file should be created exists
     if os.path.isdir(os.path.dirname(abs_file_path)):
         return abs_file_path
     else:
         raise ValueError(
-            f"The directory {os.path.commonpath(abs_file_path)} in which the file should be created does not exist."
+            f"The directory '{os.path.dirname(abs_file_path)}' in which the file should be created does not exist."
         )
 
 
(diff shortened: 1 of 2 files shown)


alphabase/io/tempmmap.py
```diff
@@ -303,25 +305,35 @@ def ones(shape: tuple, dtype: np.dtype) -> np.ndarray:
 
 
 @atexit.register
-def clear() -> str:
+def _clear() -> None:
```
GeorgWa, line 308: why is there a `_clear` and a `clear` function?

alphabase/io/tempmmap.py
```diff
@@ -303,25 +305,35 @@ def ones(shape: tuple, dtype: np.dtype) -> np.ndarray:
...
 
+    WARNING: All existing temp mmapp arrays will be unusable!
+    """
+    global _TEMP_DIR, TEMP_DIR_NAME
+
+    if _TEMP_DIR is not None:
+        logging.warning(
+            f"Folder {TEMP_DIR_NAME} with temp mmap arrays is being deleted. "
+            "All existing temp mmapp arrays will be unusable!"
+        )
```
GeorgWa, line 319: where is the delete operation of the folder?

tests/unit/io/test_tempmmap.py
```diff
@@ -0,0 +1,214 @@
...
+    import alphabase.io.tempmmap  # noqa
+
+
+def teardown_function(function):
+    """Simulate `atexit.register` and delete the module before every test."""
+    tempmmap = sys.modules["alphabase.io.tempmmap"]
+    tempmmap._clear()  # simulating @atexit.register
+
+    # # later:
+    # assert tempmmap._TEMP_DIR is None
```
GeorgWa, line 26: should this be uncommented?

GeorgWa (approved):
LGTM
this still feels like stuff which is not possible with python :D

alphabase/io/tempmmap.py
mschwoer: This is to separate the desired `atexit` behaviour (clearing only) from the functionality in `clear` (clearing + creating new) without breaking the public API

tests/unit/io/test_tempmmap.py
mschwoer: done here https://github.com/MannLabs/alphabase/pull/262/files#diff-dbc614b981464f645a25341093c73112d1e601a4902f64bc6b8bc2cd55e37df4L25-R26

alphabase/io/tempmmap.py
mschwoer: tempdir will delete it once the reference is gone
image

will add a comment

PR #265, opened 2025-01-04
Add SpecLibFlat Interface for conversion

GeorgWa:
Add new speclib flat conversion to SpecLibFlat object and remove old functions

### DIFF
--- a/alphabase/spectral_library/flat.py
+++ b/alphabase/spectral_library/flat.py
@@ -372,13 +372,8 @@ class SpecLibFlat(SpecLibBase):
             setattr(speclib_base, f"_fragment_{col}_df", df)
 
         # Drop flat indices from precursor_df if they exist
-        drop_columns = [
-            col
-            for col in ["flat_frag_start_idx", "flat_frag_stop_idx"]
-            if col in speclib_base._precursor_df.columns
-        ]
         speclib_base._precursor_df = speclib_base._precursor_df.drop(
-            drop_columns, axis=1
+            ["flat_frag_start_idx", "flat_frag_stop_idx"], axis=1, errors="ignore"
         )
 
         return speclib_base
--- a/alphabase/spectral_library/flat.py
+++ b/alphabase/spectral_library/flat.py
@@ -1,10 +1,11 @@
 import warnings
+from typing import Union
 
 import pandas as pd
 
 from alphabase.io.hdf import HDF_File
 from alphabase.peptide.fragment import (
-    _create_dense_matrices,
+    create_dense_matrices,
     flatten_fragments,
     remove_unused_fragments,
 )
@@ -250,8 +251,8 @@ class SpecLibFlat(SpecLibBase):
 
     def calc_dense_fragments(
         self,
-        additional_columns: list | None = None,
-        charged_frag_types: list | None = None,
+        additional_columns: Union[list, None] = None,
+        charged_frag_types: Union[list, None] = None,
     ) -> None:
         """
         Create a hybrid SpecLibFlat which has both flat and dense fragment representations.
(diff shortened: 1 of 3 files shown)


jalew188 (approved):
OOP LGTM

alphabase/spectral_library/flat.py
```diff
@@ -240,152 +248,134 @@ def get_full_charged_types(self, frag_df: pd.DataFrame) -> list:
...
+        ]
+        speclib_base._precursor_df = speclib_base._precursor_df.drop(
+            drop_columns, axis=1
+        )
+
+        return speclib_base
+
+    def to_SpecLibBase(self):
+        # raise a deprecation warning
+        warnings.warn(
```
mschwoer, line 377: 👍

alphabase/spectral_library/flat.py
```diff
@@ -240,152 +248,134 @@ def get_full_charged_types(self, frag_df: pd.DataFrame) -> list:
...
+
+        for col, df in df_collection.items():
+            setattr(speclib_base, f"_fragment_{col}_df", df)
+
+        # Drop flat indices from precursor_df if they exist
+        drop_columns = [
+            col
+            for col in ["flat_frag_start_idx", "flat_frag_stop_idx"]
+            if col in speclib_base._precursor_df.columns
+        ]
```
mschwoer, line 368: could be removed by using `drop(..., errors="ignore")` cf. https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html

alphabase/spectral_library/flat.py
```diff
@@ -240,152 +248,134 @@ def get_full_charged_types(self, frag_df: pd.DataFrame) -> list:
...
-        row_indices = (
-            self._fragment_df["precursor_idx"].map(
-                precursor_idx_to_accumlated_nAA, na_action="ignore"
-            )
-            + self._fragment_df["position"]
-        )
+        Creates a new SpecLibBase containing fragment_mz_df (using calculated m/z values) and
+        fragment_intensity_df by default. For each additional column specified, creates a
+        corresponding _fragment_<column>_df matrix. Including 'mz' in additional_columns will
+        use observed rather than calculated m/z values.
```
mschwoer, line 316: This behaviour (`mz`) is a bit implicit. 
could this be an extra switch (`use_observed_mz`, defaulting to `True`) ?

Could users be interested also in the calculated mz values?

alphabase/spectral_library/flat.py
```diff
@@ -1,10 +1,13 @@
 import warnings
 
-import numpy as np
 import pandas as pd
 
 from alphabase.io.hdf import HDF_File
-from alphabase.peptide.fragment import flatten_fragments, remove_unused_fragments
+from alphabase.peptide.fragment import (
+    _create_dense_matrices,
```
mschwoer, line 7: this should not be private

alphabase/spectral_library/flat.py
GeorgWa: I actually thought about this and ended up using this pattern:
https://stackoverflow.com/a/66322837

errors='ignore' felt hacky / like it could have unintended side effects

mschwoer: hm, not sure how this could have side effects..

this one would be shorter: https://stackoverflow.com/a/75367739
(and scales better in case you need to drop 10^6 columns ;-) )

mschwoer (approved):

alphabase/spectral_library/flat.py
mschwoer: btw in terms of SO upvotes, this clearly wins https://stackoverflow.com/a/59116726

GeorgWa: > hm, not sure how this could have side effects..
> 
> this one would be shorter: https://stackoverflow.com/a/75367739 (and scales better in case you need to drop 10^6 columns ;-) )

I wouldn't settle for less

PR #266, opened 2025-01-06
Remove charged frag types filtering

GeorgWa:
Add validation and soarting on loading and fixing a bit of a design flaw :D

Currently `SpecLibBase` removes frag types not in `self.charged_frag_types`.
To load them the user would need to know before loading what fragments to expect in a library.
Instead, the loading process should set sorted and valid `self.charged_frag_types` from the hdf.


tests/test_fragment.py
```diff
@@ -725,3 +726,38 @@ def test_calc_dense_fragments():
     np.testing.assert_array_equal(
         speclib_flat._fragment_correlation_df.values, expected_correlation
     )
+
+
+def test_validate_charged_frag_types():
```
mschwoer, line 731: this could be split into 6 separate tests

alphabase/spectral_library/flat.py
```diff
@@ -198,8 +200,20 @@ def load_hdf(self, hdf_file: str, load_mod_seq: bool = False):
         )
         self._fragment_df = _hdf.library.fragment_df.values
         self._protein_df = _hdf.library.protein_df.values
+
         self._fragment_mz_df = _hdf.library.fragment_mz_df.values
```
mschwoer, line 204: I would prefer storing this into a variable, and assign to `self._fragment_mz_df` only once
```
        fragment_mz_df = _hdf.library.fragment_mz_df.values
        self._fragment_mz_df = fragment_mz_df[
            sort_charged_frag_types(
                validate_charged_frag_types(fragment_mz_df.columns.values)
            )
        ]
```
(also below for `_fragment_intensity_df`)

btw I find it very confusing that `type(_hdf.library.fragment_mz_df.values)` is `pd.DataFrame` and not `np.ndarray`

Now that I know that I would "strongly prefer" (TM) to assign only once as you have two different types here

alphabase/spectral_library/base.py
```diff
@@ -717,21 +719,17 @@ def load_hdf(
...
-                frag
-                for frag in self.charged_frag_types
-                if frag in self._fragment_mz_df.columns
-            ]
+            sort_charged_frag_types(
+                validate_charged_frag_types(self._fragment_mz_df.columns.values)
+            )
         ]
+
+        self._fragment_intensity_df = _hdf.library.fragment_intensity_df.values
```
mschwoer, line 728: (same comment as belows: assign only once)

alphabase/peptide/fragment.py
```diff
@@ -346,6 +346,51 @@ def get_charged_frag_types(
     return sort_charged_frag_types(charged_frag_types)
 
 
+def validate_charged_frag_types(
```
mschwoer, line 349: from a method called `validate_..` I would not expect to return something (only raise, warn, ..) ..
`extract_valid_charged_frag_types`?

alphabase/peptide/fragment.py
```diff
@@ -346,6 +346,51 @@ def get_charged_frag_types(
...
+    charged_frag_types: List[str],
+) -> Tuple[List[str], List[bool]]:
+    """
+    Validates a list of charged fragment types and returns the valid fragment types.
+    A valid charged fragment type must:
+    1. Follow the format: {fragment_type}_z{charge} (e.g. 'b_z1', 'y_modloss_z2')
+    2. Use a fragment type that exists in FRAGMENT_TYPES
+    3. Have a positive integer charge
```
mschwoer, line 357: from this I would expect that `b_z1.2` would be filtered out, but it isn't (bc of the integer conversion in `parse_charged_frag_type`) .. so either change that or the docstring ;-)

alphabase/peptide/fragment.py
```diff
@@ -346,6 +346,51 @@ def get_charged_frag_types(
...
+    List[str]
+        List of valid charged fragment types
+
+    Examples
+    --------
+    >>> valid_types = validate_charged_frag_types(['b_z1', 'y_z2', 'invalid_z1', 'b_modloss_z2'])
+    >>> valid_types
+    ['b_z1', 'y_z2', 'b_modloss_z2']
+    """
+    import warnings
```
mschwoer, line 375: import to top?

alphabase/peptide/fragment.py
```diff
@@ -346,6 +346,51 @@ def get_charged_frag_types(
...
+
+    for frag_type in charged_frag_types:
+        base_type, charge = parse_charged_frag_type(frag_type)
+        if base_type not in FRAGMENT_TYPES:
+            warnings.warn(
+                f"Fragment type {frag_type} is not supported and will be ignored"
+            )
+            continue
+        elif charge <= 0:
+            warnings.warn(f"Charge state of fragment type {frag_type} is not positive")
```
mschwoer, line 387: `.., will be ignored`

alphabase/peptide/fragment.py
```diff
@@ -346,6 +346,51 @@ def get_charged_frag_types(
     return sort_charged_frag_types(charged_frag_types)
 
 
+def validate_charged_frag_types(
+    charged_frag_types: List[str],
+) -> Tuple[List[str], List[bool]]:
```
mschwoer, line 351: `-> List[str]`

alphabase/spectral_library/flat.py
GeorgWa: BTW, I was equally surprised that type(_hdf.library.fragment_mz_df.values) is pd.DataFrame and still don't get why...

alphabase/peptide/fragment.py
GeorgWa: added it to parse_charged_frag_type

mschwoer (approved):
LGTM

alphabase/peptide/fragment.py
```diff
@@ -379,16 +424,20 @@ def parse_charged_frag_type(charged_frag_type: str) -> Tuple[str, int]:
...
     ValueError
-        If charge state is not an integer or if fragment type is not charged
+        If charge state is not a positive integer or if fragment type is not charged
     """
-
     if charged_frag_type.count(FRAGMENT_CHARGE_SEPARATOR) == 1:
         fragment_type, charge = charged_frag_type.split(FRAGMENT_CHARGE_SEPARATOR)
         try:
+            # Check if charge is a valid integer string (no decimals)
+            if not charge.isdigit():
```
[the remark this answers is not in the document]
GeorgWa: Implemented this after a reviewer (whith a physics PhD) asked me to handle the case of fractional electrons. 😄 
```
'1.5'.isdigit() == False
int('1.5') == True
```

image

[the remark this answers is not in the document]
mschwoer: I was maybe confused by a coworker who briefly suggested to consider also the electron masses?

anyway, we're handling user input here aren't we? The Average User typically has even worse ideas when it comes to screwing up the input so I think this check is worth it.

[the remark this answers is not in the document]
GeorgWa: Absolutely 👌💯

PR #267, opened 2025-01-08
'alphabase.peptide.fragment.py' bug fix

CRYoushiwo:

### DIFF
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -928,13 +928,13 @@ def remove_unused_fragments(
         returns the reindexed precursor DataFrame and the sliced fragment DataFrames
     """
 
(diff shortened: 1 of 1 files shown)


GeorgWa:
Hi, thank you! I'll review this later.
Could you please open the Pull Request against the Development branch?
Regarding the functions in alpha base - I believe they shouldn't have any requirements about the ordering in the precursor_df. While predicted and imported libraries are typically sorted by nAA and have monotonically increasing frag_start_index, we shouldn't rely on this assumption.
I think the best approach would be to follow Feng's pattern: check if we can use an optimized function (like with monotonic nAA) and only use that function when the conditions are met.

GeorgWa:
It looks like I can't tag you. I've also proposed it here: https://github.com/MannLabs/alphabase/pull/268
Please feel free to make suggestions :)

PR #270, opened 2025-01-08
remove_unused_fragments bug fix

### LINKED ISSUE #269: remove_unused_fragments() disrupts the order of other dataframe attributes like `nAA`
**Describe the bug**

**changes**
- fix bug
- adding test and test data

CRYoushiwo:
fix #269


tests/test_remove_unused_fragments.py
```diff
@@ -0,0 +1,118 @@
+import os
+
+import alphabase.io.hdf
+from alphabase.peptide.fragment import remove_unused_fragments
+
+
+def example_remove_unused_fragments():
```
jalew188, line 7: Use pytest, see other py files in the tests folder

alphabase/peptide/fragment.py
```diff
@@ -928,13 +928,12 @@ def remove_unused_fragments(
         returns the reindexed precursor DataFrame and the sliced fragment DataFrames
     """
 
-    precursor_df = precursor_df.sort_values([frag_start_col], ascending=True)
     frag_idx = precursor_df[[frag_start_col, frag_stop_col]].values
 
     new_frag_idx, fragment_pointer = compress_fragment_indices(frag_idx)
 
     precursor_df[[frag_start_col, frag_stop_col]] = new_frag_idx
-    precursor_df = precursor_df.sort_index()
```
jalew188, line 937: sort_index() is to keep the original order of returned precursr_df as the input precursor_df. May be it is fine to keep this here.

CRYoushiwo: You're right; the original order should be preserved.

PR #271, opened 2025-01-09
remove_unused_fragments bug fix 

### LINKED ISSUE #269: remove_unused_fragments() disrupts the order of other dataframe attributes like `nAA`

CRYoushiwo:
fix #269


test_data/unit_tests/input_hdf_formats/mini_sample_remove_unused_fragments.hdf
jalew188, line 1: I think we don't need this file to make github repo cleaner. We can create precursor_df and frag_dfs on the fiy.

GeorgWa:
Please have a look here: #269 
I’m sure we can change the PR to make it compatible!

PR #276, opened 2025-01-16
clean the requirements

mschwoer:
Clean some unused dependencies

@jalew188 what about  pydivsufsort ? this is used by lcp_digest.py but was already commented?


GeorgWa (approved):
LGTM

jalew188:
pydivsufsort is not stable in some OS versions (or systems), so we did not include it as the default reqs.

requirements/requirements.txt
```diff
@@ -1,23 +1,19 @@
...
-scikit-learn==1.5.2
 regex==2024.11.6
-# pydivsufsort
+# pydivsufsort  # TODO this is used by lcp_digest.py?
 dask==2024.8.0
 dask_expr==1.1.10
 pyahocorasick==2.1.0
 pyteomics==4.7.5
-lxml==5.3.0
+lxml==5.3.0  # required by pyteomics
```
jalew188, line 18: pyteomics is not stable sometimes with errors due to `pd.version.version`. I am thinking about removing it from default reqs as it only used by AlphaRaw. What do you think @GeorgWa @mschwoer

jalew188 (approved):
LGTM

requirements/requirements.txt
GeorgWa: Can we still support mzml if we remove it?
I would not want to build and maintain our own mzml reader. But otherwise I’m happy to remove it.

jalew188: This requires by PepXML right? Not pyteomics? Otherwise we can move it into alpharaw

mschwoer: the error message requiring `lxml` came from `pepxml`, yes ..
but `pyteomics` should be stable if we fix the version, right (like `pyteomics==4.7.5`)?

on the other hand, people typically use the "loose" version (right, @GeorgWa  ;-) )

we could have a separate extra `alphabase[mzml]` (and `mzml-stable`), such that only then `pyteomics` and `lxml` will be installed?

mschwoer: made it optional

mschwoer: are we aware of any downstream packages that always require the `mzml` extra?

jalew188: I think mzml will be replaced by binary storage such as hdf5... So I prefer to make it optional. We can support it in the released installers.

mschwoer: alright, as there's little value (for now) in making `mzml` optional here, but added complexity, I chose to revert this change for now ..

PR #288, opened 2025-01-22
fix data_downloader not working when progressbar is installed

mschwoer:
My goal was to make data_downloader independent of progress bar, now it did not work anymore when progressbar was installed


ammarcsj (approved):
LGTM

PR #289, opened 2025-01-23
Refactor fdr

mschwoer:
Refactor FDR handling to prepare for extended DIANN filtering:
- do it in a dedicated method

Moving it up in the hierarchy (right after reading) is a bit tricky due to`_load_modifications` relyin on `psm_df` and `origin_df` having the same shape..
Is it worth moving it up (e.g. in terms of speed) @GeorgWa @jalew188 ?


jalew188 (approved):
LGTM

PR #294, opened 2025-02-17
allow user-defined modifications when using multiprocessing

mschwoer:
Allow to use custom modifications if multithreading is used.
Simple solution with fallback to single thread.

### DIFF
--- a/alphabase/spectral_library/base.py
+++ b/alphabase/spectral_library/base.py
@@ -7,6 +7,7 @@ import warnings
 import numpy as np
 import pandas as pd
 
+from alphabase.constants.modification import has_custom_mods
 from alphabase.io.hdf import HDF_File
 from alphabase.peptide.fragment import (
     calc_fragment_count,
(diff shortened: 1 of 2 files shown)


alphabase/spectral_library/base.py
```diff
@@ -424,7 +425,18 @@ def calc_precursor_isotope_intensity(
         if "precursor_mz" not in self._precursor_df.columns:
             self.calc_and_clip_precursor_mz()
 
-        if mp_process_num > 1 and len(self.precursor_df) > mp_batch_size:
+        do_multiprocessing = (
```
jalew188, line 428: OK, this is a temp solution...

jalew188 (approved):

alphabase/spectral_library/base.py
GeorgWa: yes, just to mitigate it crashing

GeorgWa (approved):
LGTM

PR #296, opened 2025-02-24
feat: per aa smiles encoding

mo-sameh:
Encoding a peptide to a list of smiles, one per each amino acid instead of a single smiles for the whole peptide (current implementation).

### DIFF
--- a/alphabase/smiles/peptide.py
+++ b/alphabase/smiles/peptide.py
@@ -16,7 +16,7 @@ class PeptideSmilesEncoder:
     def _parse_modifications(
         self,
         mods: Optional[str] = "",
-        mod_site: Optional[str] = "",
+        mod_sites: Optional[str] = "",
     ) -> dict:
         """
         Parse the modifications and sites into a dictionary.
@@ -25,16 +25,18 @@ class PeptideSmilesEncoder:
         ----------
         mods : Optional[str]
             Modifications in the format "GG@K;Oxidation@M;Carbamidomethyl@C".
-        mod_site : Optional[str]
+        mod_sites : Optional[str]
             Corresponding modification sites in the format "4;5;6".
 
         Returns
         -------
         dict
-            Dictionary mapping position indices to modification strings.
+            Dictionary mapping position indices to modification strings, empty dict if no mods and mod_sites provided.
         """
-        if mods and mod_site:
-            return {int(m): mod for m, mod in zip(mod_site.split(";"), mods.split(";"))}
+        if mods and mod_sites:
+            return {
+                int(m): mod for m, mod in zip(mod_sites.split(";"), mods.split(";"))
+            }
         return {}
 
     def _get_terminal_placeholders(self):
@@ -53,6 +55,24 @@ class PeptideSmilesEncoder:
             self.amino_acid_modifier.C_TERM_PLACEHOLDER, sanitize=False
         )
         return n_term_placeholder_mol, c_term_placeholder_mol
+
+    def _get_terminal_mods(self, mods: dict) -> tuple:
+        """
+        Get the N-terminal and C-terminal modifications.
+
+        Parameters
+        ----------
+        mods : dict
+            Dictionary of modifications.
+
+        Returns
+        -------
+        tuple
+            (n_term_mod, c_term_mod)
+        """
+        n_term_mod = mods.get(0)
+        c_term_mod = mods.get(-1, None)
+        return n_term_mod, c_term_mod
 
     def _get_amino_acid_smiles(self, aa: str, mod_idx: int, mods: dict) -> str:
         """
@@ -201,8 +221,7 @@ class PeptideSmilesEncoder:
         n_term_placeholder_mol, c_term_placeholder_mol = (
             self._get_terminal_placeholders()
         )
-        n_term_mod = mods.get(0, None)
-        c_term_mod = mods.get(-1, None)
+        n_term_mod, c_term_mod = self._get_terminal_mods(mods)
 
         # Process each amino acid in the sequence
         smiles_list = []
@@ -304,8 +323,6 @@ class PeptideSmilesEncoder:
         n_term_placeholder_mol, c_term_placeholder_mol = (
             self._get_terminal_placeholders()
         )
-        n_term_mod = mods.get(0, None)
-        c_term_mod = mods.get(-1, None)
 
         # Get molecules for each amino acid in the sequence
         for idx, aa in enumerate(sequence):
@@ -314,6 +331,7 @@ class PeptideSmilesEncoder:
             amino_acid_mols.append(aa_mol)
 
         # Now, assemble the peptide
+        n_term_mod, c_term_mod = self._get_terminal_mods(mods)
         peptide_mol = amino_acid_mols[0]
         peptide_mol = self.amino_acid_modifier._apply_n_terminal_modification(
             peptide_mol,


alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,94 @@ class PeptideSmilesEncoder:
...
+        c_term_placeholder_mol = Chem.MolFromSmiles(
+            self.amino_acid_modifier.C_TERM_PLACEHOLDER, sanitize=False
+        )
+
+        n_term_mod = mods.get(0, None)
+        c_term_mod = mods.get(-1, None)
+
+        # Process each amino acid in the sequence
+        for idx, aa in enumerate(sequence):
+            if idx + 1 in mods:
```
mschwoer, line 58: `idx+1` is used a lot .. -> dedicated variable?

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,94 @@ class PeptideSmilesEncoder:
...
+                aa_smiles = self.amino_acid_modifier.ptm_dict.get(mods[idx + 1], None)
+            else:
+                aa_smiles = self.amino_acid_modifier.aa_smiles.get(aa)
+            if not aa_smiles:
+                raise ValueError(
+                    f"Unknown amino acid code: {aa} or modification: {mods.get(idx + 1, 'no mod')} (SMILES for it might be missing)"
+                )
```
mschwoer, line 65: for this and the other `raise`: think about collecting all errors in a list,
then raise at the end of the method -> enables the user to fix all at once

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,94 @@ class PeptideSmilesEncoder:
...
+                    aa_mol,
+                    n_term_placeholder_mol=n_term_placeholder_mol,
+                    n_term_mod=n_term_mod,
+                )
+            else:
+                aa_mol = self.amino_acid_modifier._apply_n_terminal_modification(
+                    aa_mol,
+                    n_term_placeholder_mol=n_term_placeholder_mol,
+                    n_term_mod=None,
+                )
```
mschwoer, line 84: we are doing this for all but the first `idx`?

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,94 @@ class PeptideSmilesEncoder:
...
+
+        n_term_mod = mods.get(0, None)
+        c_term_mod = mods.get(-1, None)
+
+        # Process each amino acid in the sequence
+        for idx, aa in enumerate(sequence):
+            if idx + 1 in mods:
+                aa_smiles = self.amino_acid_modifier.ptm_dict.get(mods[idx + 1], None)
+            else:
+                aa_smiles = self.amino_acid_modifier.aa_smiles.get(aa)
```
mschwoer, line 61: (nit) for consistency, also add the `None` default here (or drop it in all the other `gets`)

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,94 @@ class PeptideSmilesEncoder:
...
+        -------
+        list
+            The list of SMILES strings for each amino acid in the peptide.
+        """
+        if mods and mod_site:
+            mods = {int(m): mod for m, mod in zip(mod_site.split(";"), mods.split(";"))}
+        else:
+            mods = {}
+
+        smiles_list = []
```
mschwoer, line 44: (nit) would put this right before the `for` loop (to make clearer it is being filled there)

tests/test_peptide_smiles.py
```diff
@@ -134,3 +134,79 @@ def test_peptide_to_mol_multiple_modifications(encoder):
...
+
+
+def test_peptide_to_smiles_per_aa_is_complete(encoder):
+    # Given a peptide sequence
+    sequence = "QMNPSIR"
+    # When converting the peptide to SMILES per amino acid
+    smiles = encoder.peptide_to_smiles_per_amino_acid(sequence)
+    # Then the number of SMILES should match the number of amino acids
+    assert isinstance(smiles, list)
+    assert len(smiles) == len(sequence)
```
mschwoer, line 146: as you have a hardcoded `sequence`: could you not just test against the actual SMILES string?

tests/test_peptide_smiles.py
```diff
@@ -134,3 +134,79 @@ def test_peptide_to_mol_multiple_modifications(encoder):
...
+
+def test_peptide_to_smiles_per_aa_no_modifications(encoder):
+    # Given a peptide sequence
+    sequence = "QMNPHIR"
+
+    formulas = [get_mod_seq_formula(aa, "") for aa in sequence]
+    expected_masses = [
+        np.sum([CHEM_MONO_MASS[elem] * n for elem, n in formula])
+        for formula in formulas
+    ]
```
mschwoer, line 157: similiar to above: could you not just set the expected masses explicitly here?
(also in rest of the tests)

tests/test_peptide_smiles.py
```diff
@@ -134,3 +134,79 @@ def test_peptide_to_mol_multiple_modifications(encoder):
...
+    generated_masses = [
+        Descriptors.ExactMolWt(Chem.MolFromSmiles(s, True)) - MASS_H2O for s in smiles
+    ]
+
+    # Then the masses should match the expected masses
+    assert np.allclose(generated_masses, expected_masses, atol=1e-4)
+
+
+def test_peptide_to_smiles_per_aa_c_terminal_modification(encoder):
+    # Given a peptide sequence and a C-terminal modification
```
mschwoer, line 193: just by symmetry: add a test for N-terminal mod?

tests/test_peptide_smiles.py
```diff
@@ -134,3 +134,79 @@ def test_peptide_to_mol_multiple_modifications(encoder):
...
+    # When converting the peptide to SMILES per amino acid
+    smiles = encoder.peptide_to_smiles_per_amino_acid(sequence)
+    # Then the number of SMILES should match the number of amino acids
+    assert isinstance(smiles, list)
+    assert len(smiles) == len(sequence)
+
+
+def test_peptide_to_smiles_per_aa_no_modifications(encoder):
+    # Given a peptide sequence
+    sequence = "QMNPHIR"
```
mschwoer, line 151: I guess parametrizing this test for some more sequences does not add any value, right? if it (doesn't) work for one, it (doesn't) work for all..?

tests/test_peptide_smiles.py
```diff
@@ -57,80 +61,171 @@ def test_peptide_to_smiles_invalid_modification(encoder):
...
-    )
+    expected_mass = 964.3837
     actual_mass = Descriptors.ExactMolWt(mol) - MASS_H2O
-    assert np.isclose(actual_mass, expected_mass, atol=1e-4)
+    assert np.isclose(
+        actual_mass, expected_mass, atol=1e-4
+    ), "Expected mass does not match actual mass"
+
+
+def test_peptide_to_smiles_per_aa_is_complete(encoder):
```
mschwoer, line 146: I think this test is obsolete now that we have `test_peptide_to_smiles_per_aa_no_modifications`?

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,217 @@ class PeptideSmilesEncoder:
...
+        ----------
+        mods : Optional[str]
+            Modifications in the format "GG@K;Oxidation@M;Carbamidomethyl@C".
+        mod_site : Optional[str]
+            Corresponding modification sites in the format "4;5;6".
+
+        Returns
+        -------
+        dict
+            Dictionary mapping position indices to modification strings.
```
mschwoer, line 34: ", empty `dict` if not both `mods` and `mod_site` are given

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,217 @@ class PeptideSmilesEncoder:
     def __init__(self):
         self.amino_acid_modifier = AminoAcidModifier()
 
+    def _parse_modifications(
+        self,
+        mods: Optional[str] = "",
+        mod_site: Optional[str] = "",
```
mschwoer, line 19: (nit) shouldn't this be `mod_sites`?

alphabase/smiles/peptide.py
```diff
@@ -13,6 +13,217 @@ class PeptideSmilesEncoder:
...
+        mod_idx = aa_idx + 1
+        aa_mol = self._get_amino_acid_mol(aa, mod_idx, mods)
+
+        aa_mol = self.amino_acid_modifier._apply_n_terminal_modification(
+            aa_mol,
+            n_term_placeholder_mol=n_term_placeholder_mol,
+            n_term_mod=n_term_mod if aa_idx == 0 else None,
+        )
+
+        aa_mol = self.amino_acid_modifier._apply_c_terminal_modification(
```
mschwoer, line 169: (nit) to avoid reusing that variable, maybe call it `aa_mol_with_n_term` in l.163, and return in l. 169?

alphabase/smiles/peptide.py
```diff
@@ -87,43 +298,19 @@ def _build_peptide(
...
-        )
-        c_term_placeholder_mol = Chem.MolFromSmiles(
-            self.amino_acid_modifier.C_TERM_PLACEHOLDER, sanitize=False
+        n_term_placeholder_mol, c_term_placeholder_mol = (
+            self._get_terminal_placeholders()
         )
-        n_term_mod = None
-        c_term_mod = None
+        n_term_mod = mods.get(0, None)
+        c_term_mod = mods.get(-1, None)
```
mschwoer, line 308: move down closer to where they are used now? l.317?

mschwoer, line 308: strictly speaking this is duplication of logic to l.204 now :-)
maybe extract into a method similar to `self._get_terminal_placeholders()`?

mschwoer (approved):
LGTM! this PR put a SMILE on my face
(by contract I am allowed to make this joke once a year)

PR #297, opened 2025-02-24
FIX: smiles to support new rdkit version

boopthesnoot:
Changed the dummy atoms to avoid them having the wrong valence


mschwoer (approved):
LGTM!

PR #301, opened 2025-02-27
Data downloader improvements

mschwoer:
- allow to download truncated versions of files. Handy for automated tests with reduced datasets.
- print size of files
- unrelated(TM): adapt handling of optional progressbar import to latest standards

### DIFF
--- a/alphabase/tools/data_downloader.py
+++ b/alphabase/tools/data_downloader.py
@@ -125,6 +125,10 @@ class FileDownloader(ABC):
             if self._is_archive:
                 self._handle_archive()
 
+            size_mb = self._get_size() / 1024**2
+            print(
+                f"{self._unzipped_output_path} successfully downloaded ({size_mb} MB)"
+            )
         else:
             size_mb = self._get_size() / 1024**2
             print(f"{self._unzipped_output_path} already exists ({size_mb} MB)")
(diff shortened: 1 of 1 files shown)


alphabase/tools/data_downloader.py
```diff
@@ -88,35 +94,80 @@ def __init__(self, url: str, output_dir: str):
...
-    def download(self) -> str:  # pragma: no cover
-        """Download file from sharing link if it does not yet exist and return its location."""
+    def download(self, max_size_kb: Optional[int] = None) -> str:  # pragma: no cover
+        """Download file from sharing link.
+
+        Parameters
+        ----------
+        max_size_kb : if given, the file will be truncated to this size in kilobytes, and the last line will be removed.
+            The file name will be appended with '.truncated{max_size_kb}'.
+            Note: don't use with binary files, as they will get corrupted!
```
ammarcsj, line 104: would it make sense to check if its a binary file and issue a warning/throw an exception?

ammarcsj (approved):
LG2M

alphabase/tools/data_downloader.py
jalew188: Or just check if TSV or CSV or TXT

mschwoer: at first glance the datashare API does not allow to do this check upfront .. and tying it to the extension might be fragile. Given that this is developer-only functionality, I would leave it as is for now.
Adding a comment to preserve your ideas.

Issue #293, opened 2025-02-16
update rdkit version


sophiamaedler:
can we upgrade the rdkit version? 

The current limitation to rdkit==2024.3.3 prevents the use of numpy2.X.   Since Numpy 1.x will be EOL in 8 months (https://scientific-python.org/specs/spec-0000/) and most packages are moving to supporting numpy2.X  I think this would make sense. 
In my packages that are dependent on alphabase, alpha base has since the update of cellpose last week become the limiting factor preventing moving to numpy2.X.

sophiamaedler:
I saw that #297 was recently merged. Does this mean we can now update rdkit?

PR #302, opened 2025-03-06
relax loose version of rdkit

mschwoer:
due to https://github.com/MannLabs/alphabase/pull/297 this should be fine now (cf. also #293)

### DIFF
--- a/requirements/requirements_loose.txt
+++ b/requirements/requirements_loose.txt
@@ -14,5 +14,5 @@ dask_expr
 pyahocorasick
 pyteomics
 lxml
-rdkit==2024.3.3 # test: tolerate_version
+rdkit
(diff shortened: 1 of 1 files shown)


requirements/requirements_loose.txt
```diff
@@ -14,5 +14,5 @@ dask_expr
 pyahocorasick
 pyteomics
 lxml
-rdkit==2024.3.3 # test: tolerate_version
+rdkit
```
jalew188, line 17: make more sense to keep rdkit up-to-date

jalew188 (approved):
make more sense to keep rdkit up-to-date

sophiamaedler (approved):

PR #303, opened 2025-03-10
Support for Custom Modifications in Multiprocessing

GeorgWa:
This PR adds support for using custom modifications in multiprocessing workflows, which was previously not supported. The key changes include:

- Added `get_custom_mods()` function to serialize user-defined modifications
- Added `init_custom_mods()` function to initialize custom modifications in worker processes
- Updated multiprocessing implementations in `calc_precursor_isotope_info_mp` and `calc_precursor_isotope_intensity_mp` to pass custom modifications to worker processes
- Removed the warning and fallback to single-process mode when custom modifications are present
- Added comprehensive unit tests for the new modification serialization functionality
- Improved documentation for multiprocessing functions
- Fixed calculation in `_count_batchify_df` for more accurate progress reporting

These changes allow users to define custom modifications and still benefit from multiprocessing performance when calculating isotope patterns, which was previously not possible.

### DIFF
--- a/alphabase/constants/modification.py
+++ b/alphabase/constants/modification.py
@@ -463,6 +463,55 @@ def has_custom_mods():
     return len(MOD_DF[MOD_DF["classification"] == _MOD_CLASSIFICATION_USER_ADDED]) > 0
 
 
+def get_custom_mods():
+    """
+    Returns a dictionary of user-defined modifications that can be serialized and passed to child processes.
+
+    Returns
+    -------
+    dict
+        Dictionary with modification names as keys and modification details as values
+    """
+    if not has_custom_mods():
+        return {}
+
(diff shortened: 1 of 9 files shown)


alphabase/constants/modification.py
```diff
@@ -463,6 +463,55 @@ def has_custom_mods():
...
+
+    Returns
+    -------
+    dict
+        Dictionary with modification names as keys and modification details as values
+    """
+    if not has_custom_mods():
+        return {}
+
+    custom_mods = MOD_DF[MOD_DF["classification"] == _MOD_CLASSIFICATION_USER_ADDED]
```
mschwoer, line 478: the main drawback I see here is that this code needs to be adapted whenever something is added to the `MOD_DF` .. this would be easily missed and then the multiprocessing part of the code would behave differently

PR #305, opened 2025-03-13
Move tests from nbs_tests/constants to pytest format

GeorgWa:
- Converted tests from Jupyter notebooks to pytest format
- Added proper fixtures to handle test isolation
- Added module-level teardown functions to restore global state

### DIFF
--- a/tests/unit/constants/test_aa.py
+++ b/tests/unit/constants/test_aa.py
@@ -1,0 +1,194 @@
+import numpy as np
+
+from alphabase.constants.aa import (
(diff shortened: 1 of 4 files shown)


GeorgWa:
This was a test for claude code. There was no manual work involved :D

mschwoer:
could we point this to `main` @GeorgWa  ?

mschwoer:
and should we remove the source notebooks? (given that they don't contain any content except those tests)

PR #312, opened 2025-03-31
311 need ccs in output table

jalew188:
See issue #311

### DIFF
--- a/alphabase/spectral_library/translate.py
+++ b/alphabase/spectral_library/translate.py
@@ -293,23 +293,11 @@ def speclib_to_single_df(
             break
     if "RT" not in df.columns:
         raise ValueError("precursor_df must contain the RT columns")
-    # if 'irt_pred' in speclib._precursor_df.columns:
-    #     df['RT'] = speclib._precursor_df['irt_pred']
-    # elif 'rt_pred' in speclib._precursor_df.columns:
-    #     df['RT'] = speclib._precursor_df['rt_pred']
-    # elif 'rt_norm' in speclib._precursor_df.columns:
-    #     df['RT'] = speclib._precursor_df['rt_norm']
-    # else:
-    #     raise ValueError('precursor_df must contain the "rt_pred" or "rt_norm" column')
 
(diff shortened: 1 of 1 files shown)


alphabase/spectral_library/translate.py
```diff
@@ -293,23 +293,16 @@ def speclib_to_single_df(
...
-    #     df['RT'] = speclib._precursor_df['rt_pred']
-    # elif 'rt_norm' in speclib._precursor_df.columns:
-    #     df['RT'] = speclib._precursor_df['rt_norm']
-    # else:
-    #     raise ValueError('precursor_df must contain the "rt_pred" or "rt_norm" column')
 
     for im_col in ["mobility_pred", "mobility"]:
         if im_col in speclib.precursor_df.columns:
             df["IonMobility"] = speclib.precursor_df[im_col]
             break
```
mschwoer, line 300: ad this pattern in general: doesn't taking the first column that is available make it hard for the client to figure out what is contained?

jalew188: For TSV lib, that is fine, users only need a value (predicted or borrowed or whatever) to be outputed in the tsv file.

alphabase/spectral_library/translate.py
```diff
@@ -293,23 +293,16 @@ def speclib_to_single_df(
...
-    # if 'mobility_pred' in speclib._precursor_df.columns:
-    #     df['IonMobility'] = speclib._precursor_df.mobility_pred
-    # elif 'mobility' in speclib._precursor_df.columns:
-    #     df['IonMobility'] = speclib._precursor_df.mobility
+
+    for ccs_col in ["ccs_pred", "ccs"]:
+        if ccs_col in speclib.precursor_df.columns:
+            df["CCS"] = speclib.precursor_df[ccs_col]
+            break
```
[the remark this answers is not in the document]
jalew188: AlphaPeptDeep always predicts ccs values, so I think the issue will be solved by this PR. And yes, ccs is a physical property with an unified unit (mobility may be not).

mschwoer (approved):

PR #313, opened 2025-04-03
Fix PSM reader unittests

lucas-diedrich:
Fix failing PSM reader tests. 

The failing tests were due to a change in the DIANN PSM reader that implemented after running the CI/CD pipeline for the last time on the new tutorial branch, but before merging the new tutorial. 

Replace the integration-like tests with real unit tests. Test coverage is not affected as equivalent integration tests exist.


tests/unit/psm_reader/test_psm_reader_provider.py
```diff
@@ -1,87 +1,30 @@
...
+        "spectronaut_report",
+        "spectronaut",
+        "sage_parquet",
+        "sage_tsv",
+    ],
+)
+def test_psm_reader_provider(reader_type):
+    """Test whether the PSM reader provider returns a PSM Reader"""
+    reader = psm_reader_provider.get_reader(reader_type=reader_type)
+    assert isinstance(reader, PSMReaderBase)
```
mschwoer, line 24: fair enough to test against `PSMReaderBase` only.. after all it's just a mapping

lucas-diedrich: I'll leave it like that for now, if that's ok

PR #314, opened 2025-04-09
NEW: annotating real spectra vs calculated one

boopthesnoot:
taken from https://github.com/MannLabs/metaptcm/blob/cbf579b52041d96636dd04c29d4855e5391bd136/metaptcm/utils/annotation.py so @mschwoer already reviewed this

### DIFF
--- a/alphabase/constants/spectral_library.py
+++ b/alphabase/constants/spectral_library.py
@@ -1,0 +1,6 @@
+"""Constants for spectral library related functionality."""
+
+# Loss type mappings - these map loss type numeric codes to their string representations
+# Used in annotate.py and flat.py
+
+LOSS_NUMBER_TO_TYPE = {0: "", 17: "_NH3", 18: "_H2O", 98: "_modloss"}
--- a/alphabase/spectral_library/annotate.py
+++ b/alphabase/spectral_library/annotate.py
@@ -1,0 +1,695 @@
+"""Module for annotating spectral libraries with raw mass spectrometry data."""
+
+from typing import Optional
+
+import numba as nb
+import numpy as np
+import pandas as pd
+from alpharaw.mzml import MzMLReader
+from alpharaw.thermo import ThermoRawData
+
+from alphabase.constants.spectral_library import LOSS_NUMBER_TO_TYPE
+from alphabase.spectral_library.flat import SpecLibFlat
+
+UNANNOTATED_TYPE = 255
+
+REQUIRED_PSM_COLUMNS = [
+    "sequence",
+    "charge",
+    "raw_name",
+    "score",
+    "proteins",
+    "spec_idx",
+    "mod_sites",
+    "mods",
+    "mod_seq_hash",
+    "mod_seq_charge_hash",
+    "flat_frag_start_idx",
+    "flat_frag_stop_idx",
+    "precursor_mz",
+]
+
+REQUIRED_RAW_COLUMNS = [
+    "_spec_idx",
+    "peak_start_idx",
+    "peak_stop_idx",
+    "precursor_mz",
+    "nce",
+    "rt",
+    "rt_max",
+    "rt_norm",
+    "activation",
+]
+
+
+def annotate_precursors_flat(
+    speclib_flat: SpecLibFlat,
+    raw_data: MzMLReader | ThermoRawData,
+    spec_idx_offset: int = 0,
+) -> SpecLibFlat:
+    """Annotate precursor information in a spectral library flat with raw data.
+
+    This function merges precursor information from the spectral library flat
+    with raw data, performs necessary data cleaning, and updates the precursor
+    DataFrame in the spectral library flat.
+
+    Parameters
+    ----------
+    speclib_flat : SpecLibFlat
+        A spectral library flat object containing precursor information.
+    raw_data : MzMLReader | ThermoRawData
+        Raw data object containing spectrum information.
+    spec_idx_offset : int, optional
+        Offset to apply to spec_idx in raw_data, by default 0.
+
+    Returns
+    -------
+    SpecLibFlat
+        An updated SpecLibFlat object with annotated precursor information.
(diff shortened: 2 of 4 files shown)


alphabase/constants/spectral_library.py
```diff
@@ -0,0 +1,6 @@
+"""Constants for spectral library related functionality."""
+
+# Loss type mappings - these map loss type numeric codes to their string representations
+# Used in annotate.py and flat.py
+
+LOSS_NUMBER_TO_TYPE = {0: "", 17: "_NH3", 18: "_H2O", 98: "_modloss"}
```
GeorgWa, line 6: This is now covered in https://github.com/MannLabs/alphabase/blob/main/alphabase/peptide/fragment.py

alphabase/spectral_library/annotate.py
```diff
@@ -0,0 +1,695 @@
...
+
+    This function merges precursor information from the spectral library flat
+    with raw data, performs necessary data cleaning, and updates the precursor
+    DataFrame in the spectral library flat.
+
+    Parameters
+    ----------
+    speclib_flat : SpecLibFlat
+        A spectral library flat object containing precursor information.
+    raw_data : MzMLReader | ThermoRawData
```
GeorgWa, line 60: I think the most accurate thig would be to use a MsDataBase from alpharaw here as annotation type.

alphabase/spectral_library/annotate.py
```diff
@@ -0,0 +1,695 @@
...
+        annotated_type,
+        annotated_loss_type,
+        annotated_charge,
+        annotated_number,
+        annotated_position,
+        annotated_error,
+    )
+
+
+def _get_dense_column(  # noqa: PLR0913
```
GeorgWa, line 374: Part of flat -> dense

alphabase/spectral_library/annotate.py
```diff
@@ -0,0 +1,695 @@
...
+    if loss_type in loss_type_mapping:
+        items.append(loss_type_mapping[loss_type])
+
+    if charge in charge_type_mapping:
+        items.append(charge_type_mapping[charge])
+
+    return "_".join(items)
+
+
+def _add_frag_column_annotation(
```
GeorgWa, line 419: Part of flat -> dense

alphabase/spectral_library/annotate.py
```diff
@@ -0,0 +1,695 @@
...
+            row["charge"],
+            frag_type_mapping,
+            loss_type_mapping,
+            charge_type_mapping,
+        ),
+        axis=1,
+    )
+
+
+def _assign_to_dense(
```
GeorgWa, line 471: Part of flat -> dense

GeorgWa (approved):
This module currently contains the conversion of a flat (annotated) library to a dense library. In the meantime we have implemented a more optimized and clean version in alphabase. 

See here: [create_dense_matrices](https://github.com/MannLabs/alphabase/blob/3e43f8a43c0320a24113af871d21df1d9404ac7f/alphabase/peptide/fragment.py#L1844)

alphabase/constants/spectral_library.py
jalew188: How much redundency is there between these two modules?

alphabase/spectral_library/annotate.py
boopthesnoot: can't use alpharaw, otherwise we get circular dependencies

alphabase/spectral_library/annotate.py
```diff
@@ -0,0 +1,557 @@
...
+            fragment_df["mz"].to_numpy(),
+            fragment_df["type"].to_numpy(),
+            fragment_df["loss_type"].to_numpy(),
+            fragment_df["charge"].to_numpy(),
+            fragment_df["number"].to_numpy(),
+            fragment_df["position"].to_numpy(),
+            mass_error_ppm=mass_error_ppm,
+        )
+
+        matched_precursor_df.loc[i, "pif"] = calculate_pif(spectrum_peak_df)
```
GeorgWa, line 206: Sorry, that I'm asking you to fix my mistakes here :D

could you change the `calulate_pif()` function so it's structured and applied in the same way as the other metrics?
So it's called after annotation and operates on the precursor df.

alphabase/spectral_library/annotate.py
```diff
@@ -0,0 +1,557 @@
...
+
+    flatlib.precursor_df["mass_accuracy"] = np.array(
+        [
+            _nan_median(flatlib._fragment_df["error"].to_numpy()[start:stop])  # noqa: SLF001
+            for start, stop in zip(start_indices, stop_indices)
+        ]
+    )
+
+
+def add_dense_lib(
```
GeorgWa, line 514: what do you think about adding this function directly to the speclib flat object?
so we can call `SpecLibFlat.add_dense_lib()`.
I think this fits the ab pattern better.

alphabase/spectral_library/flat.py
```diff
@@ -261,16 +262,16 @@ def get_full_charged_types(self, frag_df: pd.DataFrame) -> list:
         # Now if we have a fragment type that is a,b,c we should have the corresponding x,y,z
 
         corresponding = {"a": "x", "b": "y", "c": "z", "x": "a", "y": "b", "z": "c"}
-        loss_number_to_type = {0: "", 18: "_H2O", 17: "_NH3", 98: "_modloss"}
```
GeorgWa, line 264: `get_full_charged_types()` is deprecated, so there is no need to do any changes here.

alphabase/constants/spectral_library.py
GeorgWa, line 6: I think this can be removed as it's not needed by the new AB based implementation :).

GeorgWa (commented):
It's a bit integration work but then we should be good :)

PR #317, opened 2025-04-10
Move constants tests to pytest

GeorgWa:
Adapted from #305 without MP custom mods

Converted tests from Jupyter notebooks to pytest format
Added proper fixtures to handle test isolation
Added module-level teardown functions to restore global state
Removed old jupyter nb tests


jalew188 (approved):

jalew188:
Other functionalities should be moved to pytest as well.

jalew188:
> Other functionalities should be moved to pytest as well.

Nevermind, I notice that you have renamed the PR...

PR #318, opened 2025-04-10
add missing yaml dependency and tests

mschwoer:
Add the missing yaml dependency (used to come through dask)


lucas-diedrich (approved):
LGTM!

PR #331, opened 2025-08-11
Spectronaut PG Reader

lucas-diedrich:
Spectronaut protein group reader. 

Implement spectronaut protein group reader based on `pivot report` ([Spectronaut Manual](https://biognosys.com/content/uploads/2025/06/Spectronaut-20-Manual.pdf), Page 80) 

With custom post-processing method to get rid of "Filtered" values in report table.


lucas-diedrich:
Added parquet support.

PR #332, opened 2025-08-11
FragPipe PG Reader

lucas-diedrich:
Reader for FragPipe output 

Documented here: https://fragpipe.nesvilab.org/docs/tutorial_fragpipe_outputs.html#proteintsv


mschwoer (approved):
nice! the small amount of code needed is a indicator that the base class is well-designed

PR #343, opened 2025-09-16
Adapt to breaking change in datashare

mschwoer:
Adapt to the breaking changes in datashare


boopthesnoot (approved):
looks good! cool that you don't have to reupload the files

PR #344, opened 2025-09-16
new: adding diann phospho mods

boopthesnoot:
based on https://github.com/GeorgWa/alphadia-validate/pull/22#discussion_r2337737810


GeorgWa (changes requested):
I would suggest to create a separate modification mapping for DIA-NN instead of extending the maxquant mapping. Similar to msfragger.

PR #349, opened 2025-09-25
add defaults to psm_reader.yaml 

vbrennsteiner:
**What?**

To accomodate for newer spectronaut version column names, fix hardcoded charge column name in _pre_process() to instead use the configuration yaml. In particular, adding to the spectronaut/column_mapping key in the psm_reader.yaml:

- Adding 'R.Label' to 'raw_name': ['ReferenceRun']
- Adding 'FG.Charge' to 'charge': ['PrecursorCharge']

And to spectronaut/mod_seq_columns:

- Adding 'EG.ModifiedSequence'


lucas-diedrich:
superseded by #450

PR #351, opened 2025-09-27
Fix PG reader nan handling 

lucas-diedrich:
Fix incorrect parsing of missing values in some search engines (DIANN) due to the incorrect setting of the `keep_default_na` flag to `False`. This made pandas parse DIANN files with missing values as object columns with empty strings instead of float columns with nans. 

- Updated incorrect flag
- Updated reference data for DIANN integration test that incorrectly contained object dtypes.


mschwoer (approved):

mschwoer:
@lucas-diedrich seems to be the order .. not sure why that changed?
image

lucas-diedrich:
Tests pass after rerunning

PR #359, opened 2025-10-29
Support multi-level aggregation 1 | Streamline definition of PsmDfCols keys

lucas-diedrich:
Related to https://github.com/MannLabs/alphatools/pull/99#issuecomment-3462072379

> Currently, the columns do not always mean the same thing across readers. We define what each constant key actually means in the docstring of the Metaclass to have a common definition

This PR only adds **definitions** for each Key in the PsmDfCols Metaclass. In a next step, column mappings can be updated based on these standardized defintions.


alphabase/psm_reader/keys.py
```diff
@@ -20,7 +20,73 @@ def get_values(cls) -> List[str]:
...
+        Peptide sequence annotated with modification locations and standardized modification labels.
+    SEQUENCE
+        Peptide amino-acid sequence without modification annotations (plain sequence of residues).
+    DECOY
+        Label indicating whether the matched sequence is a decoy (reverse/shuffled) entry used for FDR estimation.
+    MODS
+        List of modifications on the peptide in unimod standardization
+    SCORE
+        Primary identification score for the PSM produced by the search engine quantifying quality of the PSM in arbitrary units.
+    TO_REMOVE
```
mschwoer: are all of these columns in use? should we remove the unused ones?

alphabase/psm_reader/keys.py
```diff
@@ -20,7 +20,73 @@ def get_values(cls) -> List[str]:
 
 
 class PsmDfCols(metaclass=ConstantsClass):
-    """Constants for accessing the columns of a PSM dataframe."""
+    """Constants for accessing the columns of a PSM dataframe.
+
+    Attributes
+    ----------
+    MOD_SITES
```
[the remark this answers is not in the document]
lucas-diedrich: Funnily enough, this column is not configured in any search engine, see. `psm_reader.yaml`

alphabase/psm_reader/keys.py
lucas-diedrich: Yes, will address this in another PR.

PR #366, opened 2025-11-20
Fix get cut positions

mschwoer:
Fix a bug in `get_cut_positions()` that
- would always append `len(sequence)` at the last position (even if it's already included) -> https://github.com/MannLabs/alphabase/commit/18ffb22e0869c09ebc4724fc7cb69a2add43ad65
- would have `len(sequence)+1` as the last position in case of unspecific digest -> https://github.com/MannLabs/alphabase/commit/350121ff5561ba166a4ce2761380864b4cf81ea3

### DIFF
--- a/alphabase/protein/fasta.py
+++ b/alphabase/protein/fasta.py
@@ -219,7 +219,7 @@ class Digest:
         positions = [m.start() + 1 for m in self.regex_pattern.finditer(sequence)]
         cut_pos.extend([p for p in positions if p <= len(sequence)])
 
-        if len(sequence) not in cut_pos:
+        if cut_pos[-1] != len(sequence):
             cut_pos.append(len(sequence))
 
         return np.array(cut_pos, dtype=np.int64)


alphabase/protein/fasta.py
```diff
@@ -212,10 +212,16 @@ def __init__(
...
+    def get_cut_positions(self, sequence: str) -> np.ndarray:
+        """Get cut positions for a given sequence."""
+
         cut_pos = [0]
-        cut_pos.extend([m.start() + 1 for m in self.regex_pattern.finditer(sequence)])
-        cut_pos.append(len(sequence))
+        positions = [m.start() + 1 for m in self.regex_pattern.finditer(sequence)]
+        cut_pos.extend([p for p in positions if p <= len(sequence)])
+
+        if len(sequence) not in cut_pos:
```
jalew188: if cut_pos[-1] != len(sequence)?

jalew188 (approved):
LGTM

PR #370, opened 2025-11-25
Adapt scverse tutorial to alphadia v2.0.0

lucas-diedrich:
Use `alphadia v2.0.0` instead of `DIANN` in the scverse-compatibility tutorial.


mschwoer (approved):
thanks for going this extra mile!

PR #405, opened 2026-01-27
Make numba dependency optional

mschwoer:
This would be the required changes to have the numba dependency optional. Let me know what you think.

Note that in the first two commits, I removed numba decorators. According to Claude Code, they should not have given any benefit in the first place.

Called the extra "full", as next steps could be to make `rdkit` optional. But maybe more fine-granular choice of dependencies is desired here?


README.md
```diff
@@ -96,20 +106,13 @@ be compatible with:
...
-You need to install the `hdf` extra option of the package to be able to read alphapept protein group matrices in hdf format.
-
-For those who are really adventurous, it is also possible to directly
-install any branch (e.g. `@main`) with any extras
-(e.g. `#egg=alphabase[stable,development]`) from GitHub with e.g.
-
+or
 ``` bash
-pip install "git+https://github.com/MannLabs/alphabase.git@main#egg=alphabase[stable,development]"
+pip install "alphabase[stable,full-stable]"
```
lucas-diedrich, line 111: Why does `full-stable` not automatically install `stable`?

alphabase/constants/isotope.py
```diff
@@ -54,12 +54,12 @@ def abundance_convolution(
     return ret[start:end], mono_idx
 
 
-@numba.njit
+@numba_njit
```
lucas-diedrich, line 57: The `_` is very hard to distinguish from a `.`, but I guess that's the point. Could make it `lazy_numba_njit`?

alphabase/constants/isotope.py
```diff
@@ -73,10 +73,10 @@ def one_element_dist(
     n : int
         element number.
 
-    chem_isotope_dist : numba.typed.Dict
+    chem_isotope_dist : NumbaTypedDict
```
lucas-diedrich, line 76: I think we should remove type hints from the docstrings as they tend to get outdated and sphinx can automatically infer type annotation from the type hints.

.github/workflows/pip_installation.yml
```diff
@@ -13,7 +13,7 @@ name: Pip installation and tests
 
 jobs:
   stable_installation:
-    name: Test stable pip installation on ${{ matrix.os }}
+    name: Test stable, full, ${{ matrix.os }}
```
lucas-diedrich, line 16: I think that the number of optional dependencies is getting a little hard to parse (full, etc.).

lucas-diedrich (approved):
Looks good! How large is the expected speedup by numba in the utilized scenarios?

Additional observations made by Claude:
- Add a test that tests for equivalence between `truncate_isotope` wrapped with numba and `truncate_isotope` without numba.

nbs_tests/conftest.py
```diff
@@ -0,0 +1,23 @@
+"""Shared test configuration for notebook tests."""
+
+import importlib.util
+
+import pytest
+
+NUMBA_UNAVAILABLE = importlib.util.find_spec("numba") is None
```
lucas-diedrich, line 7: Just learned about `pytest.importorskip` - functionally equivalent? 

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip

.github/workflows/pip_installation.yml
mschwoer: agree, but I prefer this added complexity over heavy dependencies polluting into packages that absoluately don't need them .. 
e.g. 
alphapepttools -> why should it have numba?
alphadia -> why should it have rdkit?

what about this:
we have very speaking names for the extras (`hdf`, `numba`, `smiles`), and one `full` extra that covers them all? I feel this combines simplicity (go with `full` if the default install does not work for you, or, if you care, install exactly what you need

(alphabase is just doing to much :-))

alphabase/constants/isotope.py
mschwoer: I see your point (no pun intended), I just lack a good name. It's actually the opposite of lazy, as it actually does something (again, no pun intended): raising in case numba is not available. Any suggestions?

nbs_tests/conftest.py
mschwoer: in this case I think no: we want to skip only certain tests in this case

.github/workflows/pip_installation.yml
lucas-diedrich: sounds good!

nbs_tests/conftest.py
lucas-diedrich: ok

alphabase/constants/isotope.py
lucas-diedrich: `conditional_numba_njit`?
but let's not make it dependent on this

README.md
mschwoer: see https://github.com/MannLabs/alphabase/pull/407

.github/workflows/pip_installation.yml
mschwoer: see https://github.com/MannLabs/alphabase/pull/407

alphabase/constants/isotope.py
mschwoer: after talking it through with Claude, I'd stick to `numba_njit` for now, but rename `numba_njit_optional -> numba_njit_or_python` ..

mschwoer:
@lucas-diedrich I did not check the numba speedup for the other parts

README.md
lucas-diedrich: nice

alphabase/constants/isotope.py
lucas-diedrich: ok, we should listen to the overlord

PR #407, opened 2026-01-28
Reduce rdkit

mschwoer:
Make rdkit dependency optional

also, rework the `full` extra which installs all optional dependencies

previous suggestion: https://github.com/MannLabs/alphabase/compare/main...smiles-extra


requirements/requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Dependencies required for running the "stable" version of alphabase.
 # Only usage of fixed versions is allowed, and all dependencies listed here must also be
 # included in `requirements_loose.txt` (enforced by a test).
-numpy<2.0 # test: tolerate_version  # rdkit==2024.3.3 is not compatible with numpy >= 2.0
+numpy<2.0 # test: tolerate_version # rdkit==2024.3.3 is not compatible with numpy >= 2.0
```
GeorgWa, line 4: what is the solution for this? Shall we bump Rdkit?

GeorgWa (approved):
Looks good

requirements/requirements.txt
lucas-diedrich: Last version that supports python 3.9 is last: 2025.9.2	

https://pypi.org/project/rdkit/

lucas-diedrich (approved):

requirements/requirements.txt
mschwoer: I would leave the numpy dep like this for now until we know what breaking changes we need to react to

lucas-diedrich: image
😭

mschwoer: at least I recall now why "hdf" was not part of "full" :-D

lucas-diedrich: my bad...

PR #416, opened 2026-06-15
Fix Windows SSL cert-store crash in data_downloader

mschwoer:
Solves troubles that occured with DataShareDownloader on windows gh runners: https://github.com/MannLabs/directlfq/actions/runs/27358174375/job/81000968038


lucas-diedrich (approved):
annoying, lgtm

PR #418, opened 2026-07-21
Conversion to and from DIA-NN .parquet

mo-sameh:
Support DIA-NN parquet spectral libraries (read & write)

- Read: `LibraryReaderBase` now loads .parquet and maps DIA-NN's dot-notation columns (via `psm_reader.yaml`).
- Write: new `translate_to_parquet` / `speclib_to_diann_df` emit DIA-NN's schema ((UniMod:N) sequences, Flags, INT64/FLOAT dtypes, no Signature); TSV path unchanged.
- Adds reader + round-trip tests and fixtures.

---

- [ ] I have read and agree to the [MannLabs Contributor License Agreement](https://github.com/MannLabs/.github/blob/main/CLA.md)


alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -723,29 +723,31 @@ library_reader_base:
   fixed_C57: False
   column_mapping:
     'raw_name': 'ReferenceRun'
-    'sequence': ['PeptideSequence', 'StrippedPeptide']
-    'modified_sequence': ['ModifiedPeptideSequence','ModifiedPeptide']
-    'charge': 'PrecursorCharge'
+    'sequence': ['PeptideSequence', 'StrippedPeptide', 'Stripped.Sequence']
```
mschwoer, line 727: the new dot notation is from the most recent DIANN version? if so, could you add the version as comment after each of the dotted fields?

otherwise, if it's only used in the DIANN speclibs, comment this also

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+    ("Exclude.From.Quant", "int"),
+    ("Protein.Ids", "str"),
+    ("Protein.Group", "str"),
+    ("Protein.Names", "str"),
+    ("Genes", "str"),
+    ("Flags", "int"),
+    ("Source.Id", "str"),
+]
+DIANN_PARQUET_COLUMN_ORDER = [name for name, _ in DIANN_PARQUET_SCHEMA]
+_DIANN_PANDAS_DTYPE = {"int": "int64", "float": "float32", "str": "str"}
```
mschwoer: `_DIANN_TO_PANDAS_DTYPE`

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+
+    Returns
+    -------
+    pd.DataFrame
+        A long-format dataframe in the DIA-NN parquet library schema.
+    """
+    if translate_mod_dict is None:
+        translate_mod_dict = mod_to_unimod_dict
+
+    if "precursor_mz" not in speclib._precursor_df.columns:
```
mschwoer: check out `PsmDfCols` -> you should be able to replace those strings with constants

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+
+    df["Modified.Sequence"] = precursor_df[["sequence", "mods", "mod_sites"]].apply(
+        create_modified_sequence,
+        axis=1,
+        translate_mod_dict=translate_mod_dict,
+        mod_sep="()",
+        nterm="",
+        cterm="",
+    )
+    df["Stripped.Sequence"] = precursor_df["sequence"]
```
mschwoer: is this creating a copy or a reference? :-)

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+        Number of precursors to convert per batch. Default: 100000
+
+    translate_mod_dict : dict
+        A dict to map AlphaBase modification names to other software.
+        Defaults to None, which uses UniMod ids, matching DIA-NN.
+    """
+    import pyarrow as pa
+    import pyarrow.parquet as pq
+
+    arrow_type = {"int": pa.int64, "float": pa.float32, "str": pa.string}
```
mschwoer: please use module-level string constants for those strings ("int", ..) coupling `arrow_type` to `DIANN_PARQUET_SCHEMA`

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+            max_frag_mz,
+        )
+    if min_frag_nAA > 0:
+        mask_fragment_intensity_by_frag_nAA(
+            speclib._fragment_intensity_df,
+            speclib._precursor_df,
+            max_mask_frag_nAA=min_frag_nAA - 1,
+        )
+
+    _speclib = SpecLibBase()
```
mschwoer: why the underscore in `_speclib` ?

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+    _speclib._fragment_intensity_df = speclib._fragment_intensity_df
+    _speclib._fragment_mz_df = speclib._fragment_mz_df
+    precursor_df = speclib._precursor_df
+
+    writer = pq.ParquetWriter(parquet_path, schema)
+    try:
+        for i in tqdm.tqdm(range(0, len(precursor_df), batch_size)):
```
mschwoer: why the batch processing? memory?

tests/integration/test_psm_readers.py
```diff
@@ -514,6 +514,17 @@ def test_diann_speclib_reader() -> None:
...
 
+def test_diann_parquet_speclib_reader() -> None:
+    """Test reading a DIA-NN 2.0 parquet spectral library."""
+    # head of a DIA-NN 2.0 `report-lib.parquet` (report-style dot-notation columns)
+    file_path = current_file_directory + "/input_data/diann_speclib_head.parquet"
+
+    reader = LibraryReaderBase()
+    reader.import_file(file_path)
+
+    _assert_reference_df_equal(reader.psm_df, "diann_parquet_speclib")
```
mschwoer, line 528: how involved would it be to test a round trip here?

alphabase/spectral_library/reader.py
```diff
@@ -243,7 +244,16 @@ def _get_fragment_intensity(self, lib_df: pd.DataFrame) -> pd.DataFrame:  # noqa
         return df
 
     def _load_file(self, filename: str) -> pd.DataFrame:
-        """Load the spectral library from a csv file."""
+        """Load the spectral library from a csv/tsv or parquet file.
+
+        DIA-NN 2.0 exports its spectral libraries as parquet (``report-lib.parquet``),
```
mschwoer: but this is our speclib class, right? can we read DIANN data into this so easily?

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
     )
 
 
+# fragment column names passed to `merge_precursor_fragment_df`
```
mschwoer: could this all be moved to a new file `translate_diann.py` ?

alphabase/spectral_library/translate.py
```diff
@@ -377,6 +378,285 @@ def speclib_to_swath_df(
...
+
+    for name, dtype in DIANN_PARQUET_SCHEMA:
+        if dtype == "str":
+            df[name] = df[name].fillna("").astype(str)
+        else:
+            df[name] = df[name].astype(_DIANN_PANDAS_DTYPE[dtype])
+    return df[DIANN_PARQUET_COLUMN_ORDER]
+
+
+def translate_to_parquet(
```
mschwoer: I am missing some link to this cool functionality from the Readme and/or docs :-)

mschwoer: also, could there be a thin wrapper in the `SpecLibBase` class that calls this method?

alphabase/spectral_library/reader.py
mschwoer: answered it myself: we can ..

this new capability should be added to docs/readme.

mschwoer (commented):
very cool!

tests/integration/test_psm_readers.py
mo-sameh: something like`test_translate_to_parquet_roundtrip` in `tests/unit/spectral_library/test_translate.py`? ;)

alphabase/spectral_library/translate.py
mo-sameh: Yes, memory. The exploded long-format fragment table require more memory than the dense SpeclibBase as the precursor cols are duplicated with each fragment

alphabase/spectral_library/translate.py
mo-sameh: It's a throwaway/local SpecLibBase used to feed each precursor batch into speclib_to_diann_df while sharing the fragment dfs.

mschwoer: in that sense every local variable is "throwaway" ;-) iirc, the convention is that local variables `_` are ignored (e.g. when returning, `val1, val2, _ = unpack_triple(xyz)`), and a leading underscore is only used to give a name to the then ignored variable `x, y, _z = unpack_triple(xyz)`

alphabase/spectral_library/translate.py
mschwoer: maybe worth a comment

tests/integration/test_psm_readers.py
mschwoer: no, more like `test_translate_to_parquet_roundtrip` in `tests/unit/spectral_library/test_translate.py` 🙈

alphabase/spectral_library/translate.py
mo-sameh: Moved it to a new `translate_diann.py`; Although I think a larger refactor for 'exporters' can be done in a separate PR, for now it imports the shared helpers from `translate.py`

alphabase/spectral_library/translate_diann.py
```diff
@@ -0,0 +1,360 @@
...
+    default: Union[str, float, None] = None,
+) -> Union[pd.Series, str, float, None]:
+    """Return the first present candidate column of `precursor_df`, else `default`."""
+    for col in candidates:
+        if col in precursor_df.columns:
+            return precursor_df[col]
+    return default
+
+
+def speclib_to_diann_df(  # noqa: PLR0913, PLR0915
```
GeorgWa, line 138: Historically the writer has been a standalone function. I think this was not an active design decision but more of a quick hack. I would favour an OOP `DIANNLibraryWriter(version='1.9')`. THoughts @mschwoer ?

alphabase/spectral_library/translate_diann.py
```diff
@@ -0,0 +1,360 @@
...
+        [PsmDfCols.SEQUENCE, PsmDfCols.MODS, PsmDfCols.MOD_SITES]
+    ].apply(
+        create_modified_sequence,
+        axis=1,
+        translate_mod_dict=translate_mod_dict,
+        mod_sep="()",
+        nterm="",
+        cterm="",
+    )
+    df["Stripped.Sequence"] = precursor_df[PsmDfCols.SEQUENCE]
```
GeorgWa: can we use constants for the diann column keys?

GeorgWa (approved):
Looks good. 
If you have Ideas how to structure the high level code of writers more OOP,
(flattening, mod trasnlation, column mapping) etc. I think this would be worth it.

alphabase/spectral_library/translate_diann.py
mschwoer: sounds good

alphabase/spectral_library/reader.py
```diff
@@ -243,7 +244,16 @@ def _get_fragment_intensity(self, lib_df: pd.DataFrame) -> pd.DataFrame:  # noqa
...
-        """Load the spectral library from a csv file."""
+        """Load the spectral library from a csv/tsv or parquet file.
+
+        DIA-NN 1.9.1+ exports its spectral libraries as parquet (``report-lib.parquet``),
+        while older DIA-NN/OpenSWATH/Spectronaut libraries are csv/tsv. Both are supported:
+        parquet files (identified by the ``.parquet`` suffix) are read with
+        :func:`pandas.read_parquet`, everything else is read as delimited text.
+        """
+        if isinstance(filename, (str, Path)) and str(filename).endswith(".parquet"):
+            return pd.read_parquet(filename)
```
mschwoer: not sure if we need the `isinstance` check here ..
downstream code would fail anyway if you passed `int` ..

either remove or make it a general check at the beginning of this method

alphabase/spectral_library/translate_diann.py
```diff
@@ -0,0 +1,360 @@
...
+    DIANN_DTYPE_FLOAT: "float32",
+    DIANN_DTYPE_STR: "str",
+}
+
+# `Flags` bitfield: bit 0 on every fragment, bit 4 on each precursor's base peak.
+_DIANN_FLAG_BASE = 1 << 0
+_DIANN_FLAG_FIRST_FRAGMENT = 1 << 4
+
+
+def _first_present(
```
mschwoer: `_get_first_present_column()`

alphabase/spectral_library/translate_diann.py
mschwoer: (could be a `TODO` and/or a separate PR)

alphabase/spectral_library/translate_diann.py
```diff
@@ -0,0 +1,360 @@
...
+    # would refine/reorder the batch, so the private frames are assigned directly here.
+    batch_speclib = SpecLibBase()
+    batch_speclib._fragment_intensity_df = speclib.fragment_intensity_df  # noqa: SLF001
+    batch_speclib._fragment_mz_df = speclib.fragment_mz_df  # noqa: SLF001
+    precursor_df = speclib.precursor_df
+
+    writer = pq.ParquetWriter(parquet_path, schema)
+    try:
+        for i in tqdm.tqdm(range(0, len(precursor_df), batch_size)):
+            batch_speclib._precursor_df = precursor_df.iloc[i : i + batch_size]  # noqa: SLF001
```
mschwoer, line 407: we update only the `_precursor_df` of the `batch_speclib`, keeping `._fragment_intensity_df` and `._fragment_mz_df` from l. 339-340.
isn't `batch_speclib` in a weird state then? maybe worth adding a comment why this works

mschwoer: (my guess is that the dangling entries of `_fragment_intensity_df` and `_fragment_mz_df` are just ignored downstream)

mschwoer (approved):

alphabase/spectral_library/translate_diann.py
mo-sameh: Good point, done

alphabase/spectral_library/reader.py
mo-sameh: This was guarding against a stream, not an `int`. `_load_file` is also called with an `io.StringIO`, which is a documented input of the base class (`PSMReaderBase._load_file` is annotated `Union[str, Path, io.StringIO]` and branches on it), used by 15 tests and several doc notebooks. Removing the check fails `test_diann_speclib_reader` and `test_msfragger_speclib_reader`.

I've kept the guard but switched it to check `io.StringIO` explicitly.

I suggest (in a separate PR): drop stream support altogether. Nothing in production uses a strea, it exists only because the tests embed 26 `TEST_DATA_*` TSV blobs as inline strings. Moving those into `tests/integration/input_data/` as fixtures would let us delete this branch, the `# for unit tests` branch in `_get_delimiter`, and rename the poorly named `filename` argument (base + 9 overrides) in one go. Let me know what you think @mschwoer ?

alphabase/spectral_library/translate_diann.py
mo-sameh: They aren't really ignored, but the full frag dfs are passed. `merge_precursor_fragment_df` indexes the fragment frames directly with each precursor's `frag_start_idx`/`frag_stop_idx`, and those are absolute offsets into the *complete* frames. We could of course slice the fragment dfs too and rebase the indices per batch, but that's extra bookkeeping for no gain, after all the point of batching here is to bound the memory needed of the *flat* output, not the library itself.

mschwoer: maybe add a brief comment explaining this

mo-sameh: Yep, already had one there 👍

PR #419, opened 2026-07-22
Add PEAKSLibraryReader for PEAKS Studio spectral libraries

sakshi327:
# PEAKSLibraryReader

Adds `PEAKSLibraryReader` to `alphabase`: reads a PEAKS Studio spectral
library export and returns an AlphaBase `SpecLibFlat`, harmonizing
sequences/modifications to UniMod and following the existing `alphabase`
reader patterns. `PEAKSLibraryReader` subclasses `PSMReaderBase` +
`SpecLibFlat` (see §2), so the usage convention matches every other library
reader: `reader = PEAKSLibraryReader(); reader.import_file("lib.tsv");
reader.precursor_df`.

**Status**: Implementation and tests complete and verified (see below).
Reviewer feedback asked to reconsider subclassing `PSMReaderBase` instead of
parsing standalone (§9 originally, now resolved) — done, and landed as 5
separate vertical-slice commits (`Slice 1`–`Slice 5` in the PR history), each
independently verified against the previous standalone implementation's
output before moving to the next. Still open as a **draft** pending further
review.

---

## 1. What was built

| File | Purpose |
|---|---|
| `alphabase/spectral_library/peaks_reader.py` | The reader itself: `PEAKSLibraryReader` class, subclassing `PSMReaderBase` + `SpecLibFlat` |
| `alphabase/constants/const_files/psm_reader.yaml` | Modified — added a `peaks:` reader-config entry and a `peaks:` entry under `modification_mappings:`, same pattern as every other reader (consumed automatically by the inherited `PSMReaderBase.__init__`, no manual wiring needed) |
| `tests/unit/spectral_library/test_peaks_reader.py` | 9 tests against small fixtures built from real file excerpts |
| `tests/unit/spectral_library/test_peaks_reader_edge_cases.py` | 36 tests against a hand-crafted edge-case fixture |
| `tests/unit/spectral_library/input_data/peaks_edge_cases.tsv` | 31-row synthetic fixture covering the edge-case list below |
| `tests/unit/spectral_library/input_data/peaks_{missing_precursor_mz,missing_charge,missing_rt,impossible_fragment,empty_library,repeated_unmapped_mod}.tsv` | 6 isolated fixtures for behaviors that must be tested alone (hard errors, empty file, repeated unmapped mod across many rows) |
| `docs/nbs/peaks_library_reader.ipynb` | User-facing tutorial notebook, same style/location as the existing `library_reader`/`library_from_fasta` tutorials, wired into the docs toctree |
| `docs/spectral_library/peaks_reader.rst` | `automodule` API-reference stub for the new module, same pattern as every other reader's docs page |

A separate, more exhaustive verification notebook (real-file run at full scale,
edge-case run, independent from-scratch manual verification, error-path
checks) exists locally but is **not** part of this PR — its results are
summarized in §7 below.

All example/reference data (`dia_db_library.tsv`, the MPIB example PEAKS
library at https://datashare.biochem.mpg.de/s/GxW5M9wmYbJW6Zz) is 52,672 rows,
897,272 fragment ions. The reader processes it end-to-end in ~1.5–2s.

## 2. Design

- `class PEAKSLibraryReader(PSMReaderBase, SpecLibFlat)` — subclasses
  `PSMReaderBase` like every other reader in `alphabase.psm_reader`, so it
  follows the standard `import_file` -> `_pre_process` -> `_translate_columns`
  -> `_load_modifications` -> `_post_process` template-method pipeline
  instead of parsing the file standalone. `__init__` chains both parents'
  `__init__` explicitly, the same style
  `alphabase.spectral_library.reader.LibraryReaderBase` uses for the
  equivalent dual inheritance. Landed as 5 separate vertical-slice commits,
  each independently verified end-to-end against the real 52,672-row file
  and/or the synthetic edge-case fixture before moving to the next:
  1. Subclass wiring + precursor-level columns (sequence/charge/precursor_mz/rt)
     via the inherited `_translate_columns`, using the existing `peaks:`
     entry in `psm_reader.yaml` — no override needed.
  2. Real modification parsing in `_load_modifications` (Unimod-name
     mapping, residue-ambiguous resolution, mass sanity check, position
     offset).
  3. `_post_process` override (own work building `fragment_df`, then
     `super()._post_process()`, same composition style as
     `alphabase.psm_reader.sage_reader.SageReaderBase`) — the first slice
     where the reader is a fully working `SpecLibFlat`.
  4. Hard-raise vs. warn-and-drop behaviors (see §4) ported into their
     correct pipeline hooks.
  5. Full fixture/real-file regression, test and tutorial-notebook updates.
- Chose `SpecLibFlat` (not the plain `SpecLibBase` that `LibraryReaderBase`
  subclasses) as the second parent, since PEAKS' fragments are flat/observed
  data (see below), not something the dense-matrix `calc_fragment_mz_df()`
  machinery `SpecLibBase` centers around applies to.
- Closest precedent for the modification-harmonization step is
  `alphabase.psm_reader.sage_reader`: Sage also reports modifications as
  (position, observed mass) rather than an embedded modified-sequence string.
- PEAKS fragments carry *observed* m/z and intensity (this is a search
  results export, not an in-silico library), so unlike `LibraryReaderBase`
  this reader never calls `calc_fragment_mz_df()` — there's nothing to
  calculate; `precursor_mz` is likewise kept as PEAKS' own observed value
  (passthrough), not recalculated from sequence+mods+charge.
- Column mapping and modification mapping are sourced from `psm_reader.yaml`
  (a `peaks:` entry, same place every other reader's config lives) via the
  standard `alphabase.psm_reader.modification_mapper.ModificationMapper`
  instance that `PSMReaderBase.__init__` builds automatically — no manual
  construction needed now that this is a real subclass.

### Position indexing for modifications
PEAKS reports modification positions as a **0-based index into the sequence
string**. Verified against **all 8,094 modification occurrences** in the real
52,672-row file (0 mismatches) by confirming the residue at that index
actually matches the modification's chemical target (e.g. Carboxymethyl only
ever lands on `C`, Oxidation only on `M`).

AlphaBase's own `mod_sites` convention is **1-based** for side-chain
modifications, so: **AlphaBase site = PEAKS position + 1**. N-terminal
modifications are always reported at PEAKS position 0 and map to AlphaBase's
fixed site `"0"` directly (not `0 + 1`).

### Modification vocabulary
Only **3 distinct (name, mass) combinations** exist in the real example file
— cross-checked against `alphabase/constants/const_files/modification.tsv`:

| PEAKS name | Mass (PEAKS-reported) | AlphaBase name | Exact UniMod mass |
|---|---|---|---|
| `Carboxymethyl` | 58.01 | `Carboxymethyl@C` | 58.005479 |
| `Oxidation (M)` | 15.99 | `Oxidation@M` | 15.994915 |
| `Acetylation (Protein N-term)` | 42.01 | `Acetyl@Protein_N-term` | 42.010565 |

**Note**: `Carboxymethyl` and `Carbamidomethyl` are easy to confuse — they are
different modifications (58.005 vs 57.021 Da). Confirmed this file uses
Carboxymethyl specifically.

**This mapping is very likely incomplete for PEAKS libraries in general** —
it's only what appears in the one example file available. An unmapped name
warns and drops the affected precursor rather than crashing (see §5), so the
reader degrades gracefully as new modifications are encountered. Extending
it is a one-line addition to the `peaks:` block in `psm_reader.yaml` (no code
change) once real examples of other modifications are available.

### Fragment ion vocabulary
Verified against **all 897,272 fragment tokens** in the real file: only `b`
and `y` ion types occur, and only `H2O`/`NH3` neutral losses — 0 unparseable
tokens in the real data.

## 4. Error-handling behavior (deliberately not uniform — each case reasoned through separately)

| Case | Behavior | Reasoning |
|---|---|---|
| Blank `sequence`, `charge`, `precursor_mz`, or `rt` | **Raises `ValueError`**, names the raw PEAKS column and row | Corrupt/unidentifiable row — stop immediately with a clear message rather than let it surface later as a cryptic pandas/numpy dtype-cast error. `rt` was added to this list after reviewer feedback pointed out the original implementation treated it as optional while the other three were hard-required; now validated in `_pre_process`, before column translation, so the error names the actual TSV header (e.g. `'m/z'`) rather than the AlphaBase column name |
| Fragment number impossible for the peptide's length (e.g. `y10` on a 5-residue peptide) | **Raises `ValueError`**, names the token | `position` is stored as an unsigned int; a negative value would otherwise silently wrap around to a huge number instead of erroring — a crash is strictly better than corrupt-but-plausible-looking data here |
| Unmapped modification name | **Warns + drops that one precursor** | Recoverable/extensible (add a mapping and re-run) — one unrecognized modification shouldn't abort a 50k-row import |
| Non-standard sequence character (lowercase, digit, punctuation) | **Warns + drops that one precursor** | Same reasoning as unmapped modification. Validated against AlphaBase's own recognized AA alphabet (not just the standard 20 — also permits `U`/`O`/`X`/`B`/`J`/`Z`, which AlphaBase itself defines as intentional extended/placeholder codes) |
| Garbled/unparseable fragment ion label within an otherwise-good precursor | **Warns + skips that one token**, keeps the precursor and its other fragments | A single bad peak in an otherwise good spectrum shouldn't cost the whole precursor |
| Header-only (empty) file | **Returns a valid empty `SpecLibFlat`**, no error | Zero rows is not an error condition |
| Two modifications at the exact same position | Both kept, no dedup, no error | Malformed input, but a defined and tested behavior rather than an unhandled exception |
| Duplicate fragment entries within one spectrum | Both kept, no dedup | Same reasoning |
| Exact duplicate precursor rows | Both kept as separate rows, no dedup | Same reasoning |
| Modifications listed out of position order | Output preserves input token order (not re-sorted by position) | Documented, tested behavior |

## 5. Bugs found and fixed while building the edge-case tests

Building the synthetic fixture surfaced 5 real bugs that the real 52,672-row
file's "happy path" data never exercised:

1. **Empty `Peaks List` crashed the whole import.** `"".split(";")` returns
   `['']`, not `[]` — that lone empty string failed to match the fragment
   regex and raised. Fixed: an empty cell now yields zero fragments.
2. **A chemically-impossible fragment number silently corrupted data instead
   of crashing.** `position` is `uint32`; a negative value (ion number
   exceeding peptide length) would wrap around to a huge positive number
   rather than erroring. Fixed: now raises a clear `ValueError` naming the
   exact token.
3. **One garbled ion label used to abort the entire precursor's fragments**
   (and originally the whole file). Changed to: warn, skip just that token,
   keep everything else.
4. **A blank `m/z` or `charge` would eventually crash several steps later**
   with a cryptic numpy dtype-cast error. Now validated immediately with a
   message naming the exact column and row.
5. **No validation of sequence characters existed at all.** Added a check
   against AlphaBase's own recognized AA alphabet; non-standard characters
   now drop the row with a warning instead of silently passing through.

All 5 fixes were regression-checked against the real 52,672-row file
afterward — identical output (same shape, same row count) before and after,
confirming none of this changed behavior on well-formed data.

A 6th bug surfaced during the subclassing rewrite (§2, slice 3): the first
draft of the `_post_process` override built `fragment_df` *before* the
inherited pipeline dropped unmapped-mod/bad-sequence precursors, so fragments
belonging to a dropped precursor were left behind, unreferenced, in
`fragment_df` — 350 fragment rows produced against the edge-case fixture
instead of the expected 302. Caught immediately by the same
equivalence-against-the-standalone-implementation check used throughout the
rewrite; fixed by filtering to surviving rows before building fragments,
matching the standalone implementation's original ordering.

## 6. Edge cases covered by the synthetic fixture (31 rows + 6 isolated)

**Modifications**: no modification (confirmed empty string, not `"nan"`
text) · single modification · 3 modifications on one peptide · N-terminal
modification · modification on the very last residue (off-by-one check:
site must equal `len(sequence)` exactly) · two modifications at the same
position · modifications listed out of position order · trailing stray
delimiter · unrecognized/unmapped modification name · residue-ambiguous
modification name resolved via the sequence (`"Deamidation (NQ)"` ->
`Deamidated@N` or `Deamidated@Q` depending on which residue is actually at
the reported position) · known modification name with a mass that doesn't
match the expected mass (dropped, not silently trusted on name alone) ·
repeated unmapped modification across many rows logs once, not once per row.

**Sequence**: very short peptide (5 aa) · very long peptide (30 aa) · all 20
standard amino acids collectively represented · invalid characters
(lowercase, digit).

**Charge**: range 1–6 · same peptide+mods observed at multiple charge states
(confirmed as separate, independently-indexed rows, not merged).

**Fragment parsing**: bare ion (no loss, no charge) · charge-only suffix ·
loss-only suffix · loss+charge combined suffix (not present in the real
file — hand-built specifically for this) · empty Peaks List (0 fragments) ·
single-fragment list · duplicate fragment entries · fragment number
exceeding peptide length · intensity extremes (exactly 0.0 and exactly 1.0).

**Retention time**: RT = 0 (confirmed distinct from missing/NaN).

**File/row level**: header-only empty library · missing required field
(blank `m/z`, blank `charge`, blank `rt`) · exact duplicate row.

**Structural**: 0/1/many-fragment precursors, confirming
`flat_frag_start_idx`/`flat_frag_stop_idx` line up exactly with no gaps or
overlaps across the whole table — the single easiest place for an
off-by-one bug to hide silently.

## 7. Verification methodology

Three independent layers, each catching a different class of bug:

1. **Unit tests** (45 total: 9 + 36) — pytest, exact-value assertions,
   `alphabase`'s existing test conventions. All passing; full `alphabase`
   unit suite (276 tests total) still passes, confirming nothing else broke.
2. **Independent manual verification** (in the notebook) — expected values
   for every field of every fixture row computed with a *from-scratch*
   parser using plain string methods (`.split()`/`.partition()`), not the
   reader's own regex, so a bug in the regex can't "verify itself." Result:
   **0 mismatches across 28 precursor-level checks and 278 fragment-level
   checks.** Exported to two CSVs for spreadsheet-based manual spot-checking.
3. **Real-file regression** — full 52,672-row file re-run after every code
   change; identical `precursor_df`/`fragment_df` shape throughout.

## 8. Known limitations

- **Modification vocabulary (3 entries) is almost certainly incomplete** —
  it's everything in the one example file, not a general PEAKS vocabulary.
  Need more example libraries with diverse modifications to extend it
  confidently.
- **C-terminal modification handling is implemented but unverified** — no
  C-terminal example exists in the available data, so that code path
  (site `"-1"`, mirroring the `sage_reader` convention) has never been
  exercised against real PEAKS output.
- Decoys are not handled (not present in the example library format at all —
  unclear whether PEAKS libraries ever carry decoy annotations).
- ~~`rt_unit="second"` is currently the only supported value~~ **Resolved**
  by the subclassing rewrite: `rt_unit` now goes through `PSMReaderBase`'s
  own inherited RT handling, which already supports
  `"minute"`/`"second"`/`"irt"` correctly, so the standalone implementation's
  artificial `NotImplementedError` restriction was removed rather than
  reimplemented.

## 9. Open questions

1. Do other PEAKS example libraries exist (more modifications, C-terminal
   mods, decoys) to validate/extend the mapping against?
2. ~~Is a standalone class (not subclassing `LibraryReaderBase`) the right
   call given the shape mismatch, or is reworking the base class hierarchy
   to accommodate "precursor-row-format" libraries preferred long-term?~~
   **Resolved**: subclasses `PSMReaderBase` + `SpecLibFlat` directly (§2),
   not `LibraryReaderBase` (whose `SpecLibBase` parent centers on dense
   in-silico fragment matrices this reader has no use for).
3. Does the hard-raise vs. warn-and-drop split in §4 match how the team
   wants error handling to work across readers generally, or is there an
   existing convention I should align to instead? (Partially addressed:
   `rt` joined the hard-raise list per reviewer feedback; the rest of the
   split is unchanged.)


tests/unit/spectral_library/test_peaks_reader.py
```diff
@@ -0,0 +1,200 @@
...
+from alphabase.spectral_library.flat import SpecLibFlat
+from alphabase.spectral_library.peaks_reader import PEAKSLibraryReader
+
+_HEADER = (
+    "m/z\tz\trt (seconds)\tActivation Mode\tSequence (backbone)\t"
+    "Modifications\tPeaks Count\tPeaks List\tEngine"
+)
+
+# unmodified peptide, single charge fragments only
+_ROW_NO_MOD = (
```
mschwoer, line 29: some test data is stored in `tsv`, others given as a string. Is there a reason for that?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+from alphabase.psm_reader.psm_reader import psm_reader_yaml
+from alphabase.spectral_library.flat import SpecLibFlat
+from alphabase.utils import _get_delimiter
+
+logger = logging.getLogger(__name__)
+
+# The set of single-letter codes AlphaBase itself recognizes (standard 20 AAs
+# plus extended codes it defines on purpose, e.g. U=selenocysteine,
+# O=pyrrolysine, X/B/J/Z=ambiguous/unknown placeholders with a disabling mass
+# - see alphabase/constants/const_files/amino_acid.tsv). Anything *outside*
```
mschwoer, line 24: "Anything.." this references a downstream implementation that uses this constant, right?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+# Very likely incomplete for PEAKS libraries in general - extend the yaml
+# entry as new modifications are encountered (an unmapped modification warns
+# and drops the affected precursor rather than crashing - see
+# `_harmonize_modifications`).
+
+# Matches one modification token in PEAKS' "Modifications" column, e.g.
+#   "10-Carboxymethyl-(58.01)" -> position="10", name="Carboxymethyl", mass="58.01"
+#   "0-Acetylation (Protein N-term)-(42.01)" -> position="0", name="Acetylation (Protein N-term)"
+# Multiple modifications are concatenated with ';' in the source column, but
+# this pattern is applied with `finditer` (not `split`), so it finds every
```
mschwoer: again, here the implemtation should not be referenced (does not age well).. cf. https://github.com/MannLabs/alphabase/pull/419/changes#r3630693682

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+    -------------------------------------------
+    - `alphabase.spectral_library.reader.LibraryReaderBase` is the existing
+      spectral library reader, but it expects "long format" input (one row
+      per *fragment*, like Spectronaut exports). PEAKS' format is the
+      opposite shape (one row per *precursor*), so this class does not
+      subclass it - it builds `precursor_df`/`fragment_df` directly instead.
+    - `alphabase.psm_reader.sage_reader.SageReaderBase` is the closest
+      precedent for the modification-harmonization step: Sage also reports
+      modifications as (position, observed mass) rather than an embedded
+      modified-sequence string.
```
mschwoer: these details might be valuable during implementation, but not for the reader of this docstring ;-)

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+        modification_mapping : dict, optional
+            Additional/override PEAKS-name -> AlphaBase-name mappings, merged
+            on top of the `"peaks"` entry in `psm_reader.yaml`'s
+            `modification_mappings` section. See
+            `alphabase.psm_reader.modification_mapper.ModificationMapper`.
+
+        rt_unit : str, optional
+            Unit of the retention time column. Defaults to
+            `psm_reader_yaml["peaks"]["rt_unit"]` ("second" - the only
+            convention observed in PEAKS exports so far).
```
mschwoer: "so far" ?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+        raw_df = raw_df.loc[keep_mask].reset_index(drop=True)
+
+        precursor_df[PsmDfCols.NAA] = precursor_df[PsmDfCols.SEQUENCE].str.len()
+
+        # AlphaBase's internal `rt` convention is minutes (mirrors
+        # PSMReaderBase._normalize_rt's handling of rt_unit == "second"). RT
+        # is not a required field (unlike sequence/charge/precursor_mz): a
+        # blank/unparseable value becomes NaN rather than raising, since RT
+        # isn't needed to identify a precursor.
+        precursor_df[PsmDfCols.RT] = (
```
mschwoer: shouldn't `RT` also be in `_REQUIRED_COLUMNS` then?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+            Library with `precursor_df` (one row per precursor) and
+            `fragment_df` (one row per fragment ion), linked via the
+            `flat_frag_start_idx` / `flat_frag_stop_idx` columns.
+
+        """
+        raw_df = self._load_file(filename)
+
+        precursor_df, raw_df = self._parse_precursors(raw_df)
+        precursor_df, raw_df = self._harmonize_modifications(precursor_df, raw_df)
+        precursor_df, fragment_df = self._parse_fragments(precursor_df, raw_df)
```
mschwoer: what was the reason to not adopt the pattern (storing dfs as instance variables) in `PSMReaderBase.import_file()`? Would make transferring this to a subclass easier later on

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+                    continue
+
+                peaks_position = int(match.group("position"))
+                site = self._mod_site(alphabase_name, peaks_position)
+
+                names.append(alphabase_name)
+                sites.append(site)
+
+            if unknown:
+                logger.warning(
```
mschwoer: won't this log on every precursor that has an unknown mod? -> lots of noise

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+                stacklevel=2,
+            )
+
+        return (
+            precursor_df.loc[keep_mask].reset_index(drop=True),
+            raw_df.loc[keep_mask].reset_index(drop=True),
+        )
+
+    @staticmethod
+    def _mod_site(alphabase_mod_name: str, peaks_position: int) -> str:
```
mschwoer: (nit) `_convert_mod_site`

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+            # An empty "Peaks List" (Peaks Count == 0) means zero fragments,
+            # not one token: "".split(";") would otherwise yield [""], which
+            # doesn't match the fragment regex and would be treated as a
+            # (spuriously) unparseable token.
+            tokens = peaks_list.split(";") if peaks_list else []
+
+            for token in tokens:
+                match = _PEAKS_FRAGMENT_TOKEN_RE.match(token)
+                if match is None:
+                    logger.warning(
```
mschwoer: log noise? cf https://github.com/MannLabs/alphabase/pull/419/changes#r3630768747

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,494 @@
...
+    r"(?P<ion_type>[by])(?P<ion_number>\d+)"
+    r"(?P<neutral_loss>-(?:H2O|NH3))?"
+    r"(?:\[(?P<frag_charge>\d+)\+\])?$"
+)
+
+_PROTEIN_N_TERM_SITE = "0"
+_C_TERM_SITE = "-1"
+
+
+class PEAKSLibraryReader:
```
mschwoer: we discussed this briefly in person, please consider subclassing `PSMReaderBase`.. otherwise this is disconnected from the rest of the readers and will drift ..

tests/unit/spectral_library/test_peaks_reader.py
sakshi327: Yes — it's a split by purpose rather than an inconsistency:

- **Inline strings** (`test_peaks_reader.py`): each row is a small, hand-picked example isolating one specific parsing branch (single mod, multi-mod, fragment charge suffix, etc.), so keeping the row text right next to the assertion that explains it makes the test self-contained and easy to read top-to-bottom without jumping to another file.
- **`.tsv` fixtures** (`test_peaks_reader_edge_cases.py` + the four `peaks_missing_*`/`peaks_impossible_fragment`/`peaks_empty_library` files): these are either a large 31-row file reused across ~25 tests (too big to inline readably) or single-purpose files for a hard-error condition that can't coexist with a successful import in the same file.

Happy to consolidate if you'd prefer one convention throughout — let me know which way you'd lean.

alphabase/spectral_library/peaks_reader.py
sakshi327: Done — see the 5-commit vertical slice starting at 'Slice 1: subclass PSMReaderBase...'. Reader now follows the standard import_file→_pre_process→...→_post_process pipeline like the rest of psm_reader.

alphabase/spectral_library/peaks_reader.py
sakshi327: Moved that reasoning into a plain # comment above the constant it explains, out of the docstring.

alphabase/spectral_library/peaks_reader.py
sakshi327: Agreed; implementation rationale lives in code comments now.

alphabase/spectral_library/peaks_reader.py
sakshi327: Yes, agreed, it's hard-required now (_pre_process raises on a blank value), matching sequence/charge/precursor_mz.

alphabase/spectral_library/peaks_reader.py
sakshi327: Fixed now; uses self._psm_df/self._precursor_df via inheritance directly, no more standalone parsing.

alphabase/spectral_library/peaks_reader.py
sakshi327: No downstream implementation reference intended; it's describing the constant itself (_VALID_SEQUENCE_CHARS, built from aa_formula.index): "anything outside this set" just means any character not in AlphaBase's recognized amino-acid alphabet. Happy to reword if it still reads as implementation-referencing.

alphabase/spectral_library/peaks_reader.py
sakshi327: Makes sense, fixed now; unmapped/mass-mismatched modifications are now collected across the whole file and logged once at the end (_load_modifications), not per precursor.

alphabase/spectral_library/peaks_reader.py
sakshi327: Same fix as the other log-noise comment; now aggregated once per file. Unparseable fragment tokens (_parse_fragments) got the same treatment.

alphabase/spectral_library/peaks_reader.py
sakshi327: Done; renamed from _mod_site to _convert_mod_site.

alphabase/spectral_library/peaks_reader.py
sakshi327: Deliberate hedge, not filler; it flags that "second" is only what's been observed in the one example library this reader was built against, not a guarantee about PEAKS exports in general. Happy to drop it if you'd rather state it plainly.

mschwoer: this should not reference the example used for building it, but rather be guarded (i.e. raise) against the unsupported cases

alphabase/spectral_library/peaks_reader.py
mschwoer: while it's a valid strategy in this case to do the slices as commits (typically you would stack several branches/PRs), in this case it's not very helpful as [9f03b70](https://github.com/MannLabs/alphabase/pull/419/commits/9f03b70a3f14f67d996829f233fe5c77cf25a52b) is the diff against the previous implementation, and not the current `main` branch..
also, typically slices come with the respective tests, such that one can review the CL as a whole

alphabase/spectral_library/peaks_reader.py
sakshi327: Fair — dropped the reference to the example library entirely. Docstring now just states the convention plainly and points at the existing guard: `_pre_process` already raises immediately if the expected rt column isn't found, rather than silently misinterpreting values in an unsupported unit. See 7ac154f.

alphabase/spectral_library/peaks_reader.py
sakshi327: Ah, I see the confusion now, sorry — I misread your comment as being about diffing against main in general, not specifically that 9f03b70's diff is against the now-abandoned standalone implementation rather than being a clean addition on top of main. That's a fair complaint: main has no peaks_reader.py at all (this whole file is new since branch point 9e80cf4), so the "old standalone implementation" commits (e34dc8f, a955096, de8cf37, c416887) are themselves unmerged PR history, not something the reader has any reason to diff against.

Proposed fix, before I touch history again — rebase the whole branch onto main and drop the standalone-implementation era from history entirely, so the file only ever exists in its subclass form. Re-sliced this way, each commit adds fresh code on top of main plus the tests for that code:

- Slice 1: `PEAKSLibraryReader(PSMReaderBase, SpecLibFlat)` skeleton — precursor-level columns only (sequence/charge/precursor_mz/rt) via `_pre_process`/`_translate_columns`, + the precursor-column-harmonization and required-column-validation tests.
- Slice 2: modification harmonization (`_load_modifications`, `_parse_mod_cell`, residue-ambiguous resolution, mass sanity check), + the modification tests.
- Slice 3: `fragment_df` construction via the Sage-style `_post_process` override, + the fragment-explosion/charge/neutral-loss tests.
- Slice 4: hard-raise-vs-warn-and-drop hardening across the edge cases (blank required fields raise; bad sequence/unmapped mod/impossible fragment warn-and-drop), + the remaining edge-case tests.
- Slice 5: notebook + real-library verification.

Each slice's diff would then be purely additive against main, and reviewable together with its own tests. Want to confirm this is the structure you'd find useful before I do the rebase and force-push.

PR #421, opened 2026-08-02
PEAKStudio Empirical/Spectral library reader

mo-sameh:
## Add PeaksLibraryReader for PEAKS Studio DIA spectral libraries

Imports a PEAKS Studio DIA spectral library export (`dia_db_library.tsv`) into a
standard AlphaBase spectral library (`precursor_df` + `fragment_intensity_df` +
`fragment_mz_df`).

- Unpacks the PEAKS wide format (fragments packed in one `Peaks List` cell) into
  the long format `LibraryReaderBase` already consumes — no core changes.
- Parses b/y fragment annotations up to charge 3 with H2O/NH3 losses.
- Resolves the `Modifications` column by mass + the residue read from the sequence;
  unmapped mods warn and drop the precursor instead of aborting.

```python
# Example use case
reader = PeaksLibraryReader()
reader.import_file("peak_dia_db_library.tsv")
```
---

- [ ] I have read and agree to the [MannLabs Contributor License Agreement](https://github.com/MannLabs/.github/blob/main/CLA.md)

### DIFF
--- a/alphabase/constants/const_files/psm_reader.yaml
+++ b/alphabase/constants/const_files/psm_reader.yaml
@@ -763,7 +763,6 @@ peaks_library:
     'precursor_mz': 'm/z'
     'rt': 'rt (seconds)'
     '_tmp_mods': 'Modifications'   # raw mod column, parsed in code (like MSFragger)
-    'peaks_list': 'Peaks List'     # packed fragment column, exploded in _load_file
     'fragment_mz': 'FragmentMz'
     'fragment_intensity': 'FragmentIntensity'
     'fragment_type': 'FragmentType'
@@ -772,6 +771,8 @@ peaks_library:
     'fragment_loss_type': 'FragmentLossType'
   mod_seq_columns:
     - 'Sequence (backbone)'
+  # packed per-precursor fragment column.
+  peaks_list_column: 'Peaks List'
   modification_mapping_type: null  # modifications resolved in code using mass and residue/terminus, not by name
   mod_mass_tol: 0.1
   # alphabase mod names the PEAKS "Modifications" column resolves to (matched by
--- a/alphabase/constants/const_files/psm_reader.yaml
+++ b/alphabase/constants/const_files/psm_reader.yaml
@@ -751,6 +751,7 @@ library_reader_base:
 # PEAKS Studio DIA library export (wide format: one row per precursor, all
 # fragments packed into a single "Peaks List" cell). PeaksLibraryReader explodes
 # it into the long format below before the standard library pipeline runs, and converts it to the dense SpeclibBase format.
+# Columns tested against PEAKS Studio 13.5.
 
 peaks_library:
   reader_type: peaks_library
--- a/alphabase/constants/const_files/psm_reader.yaml
+++ b/alphabase/constants/const_files/psm_reader.yaml
@@ -748,6 +748,38 @@ library_reader_base:
     - 'FullUniModPeptideName'
   modification_mapping_type: 'maxquant'
 
+# PEAKS Studio DIA library export (wide format: one row per precursor, all
+# fragments packed into a single "Peaks List" cell). PeaksLibraryReader explodes
+# it into the long format below before the standard library pipeline runs, and converts it to the dense SpeclibBase format.
+
+peaks_library:
+  reader_type: peaks_library
+  rt_unit: second
+  fixed_C57: False
+  column_mapping:
+    'sequence': 'Sequence (backbone)'
+    'modified_sequence': 'Sequence (backbone)'
+    'charge': 'z'
+    'precursor_mz': 'm/z'
+    'rt': 'rt (seconds)'
+    'fragment_mz': 'FragmentMz'
+    'fragment_intensity': 'FragmentIntensity'
+    'fragment_type': 'FragmentType'
+    'fragment_charge': 'FragmentCharge'
+    'fragment_series': 'FragmentNumber'
+    'fragment_loss_type': 'FragmentLossType'
+  mod_seq_columns:
+    - 'Sequence (backbone)'
+  modification_mapping_type: null  # modifications resolved in code using mass and residue/terminus, not by name
+  mod_mass_tol: 0.1
+  # alphabase mod names the PEAKS "Modifications" column resolves to (matched by
+  # mass + residue/terminus). A modification not listed here warns and its
+  # precursor is dropped; add it here to keep those precursors.
+  mass_mapped_mods:
+    - 'Carboxymethyl@C'         # 58.01
+    - 'Oxidation@M'             # 15.99
+    - 'Acetyl@Protein_N-term'   # 42.01
+
 sage:
   reader_type: sage
   rt_unit: second
--- a/alphabase/constants/const_files/psm_reader.yaml
+++ b/alphabase/constants/const_files/psm_reader.yaml
@@ -762,6 +762,8 @@ peaks_library:
     'charge': 'z'
     'precursor_mz': 'm/z'
     'rt': 'rt (seconds)'
+    '_tmp_mods': 'Modifications'   # raw mod column, parsed in code (like MSFragger)
+    'peaks_list': 'Peaks List'     # packed fragment column, exploded in _load_file
     'fragment_mz': 'FragmentMz'
     'fragment_intensity': 'FragmentIntensity'
     'fragment_type': 'FragmentType'
--- a/alphabase/constants/modification.py
+++ b/alphabase/constants/modification.py
@@ -62,6 +62,12 @@ class ModificationKeys(metaclass=_ConstantsClass):
     #: Protein C-terminus only
     PROTEIN_C_TERM: str = "Protein_C-term"
 
+    #: Suffix shared by all N-terminal sites ("Any_N-term", "Protein_N-term")
+    N_TERM: str = "N-term"
+
+    #: Suffix shared by all C-terminal sites ("Any_C-term", "Protein_C-term")
+    C_TERM: str = "C-term"
+
     #: AA-specific N-term suffix: "Gln->pyro-Glu@Q^Any_N-term" (Q at any N-term)
     ANY_N_TERM_SPECIFIC: str = "^Any_N-term"
 
--- a/alphabase/spectral_library/peaks_reader.py
+++ b/alphabase/spectral_library/peaks_reader.py
@@ -13,22 +13,9 @@ import numpy as np
 import pandas as pd
 
 from alphabase.constants.modification import MOD_MASS, ModificationKeys
-from alphabase.psm_reader.keys import PsmDfCols
+from alphabase.psm_reader.keys import LibPsmDfCols, PsmDfCols
 from alphabase.psm_reader.psm_reader import psm_reader_yaml
 from alphabase.spectral_library.reader import LibraryReaderBase
-
-# Column names in the raw PEAKS export.
-_PEAKS_SEQUENCE_COLUMN = "Sequence (backbone)"
-_PEAKS_MOD_COLUMN = "Modifications"
-_PEAKS_PEAKS_COLUMN = "Peaks List"
-
-# Internal (synthesized) long-format fragment column names, mapped in psm_reader.yaml.
-_FRAGMENT_MZ = "FragmentMz"
-_FRAGMENT_INTENSITY = "FragmentIntensity"
-_FRAGMENT_TYPE = "FragmentType"
-_FRAGMENT_CHARGE = "FragmentCharge"
-_FRAGMENT_NUMBER = "FragmentNumber"
-_FRAGMENT_LOSS_TYPE = "FragmentLossType"
 
 # A fragment annotation, e.g. 'b6', 'y13[2+]', 'b12-H2O', 'y5-NH3[3+]'.
 _FRAGMENT_ANNOTATION_PATTERN = re.compile(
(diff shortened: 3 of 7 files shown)


alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+
+    """
+    match = _FRAGMENT_ANNOTATION_PATTERN.match(annotation)
+    if match is None:
+        return None
+    frag_type, number, loss, charge = match.groups()
+    return frag_type, int(number), int(charge) if charge else 1, loss or ""
+
+
+def _match_mod_by_mass(
```
GeorgWa: Is there a way how we can reuse the mod annotation from other readers? maybe the sage reader, which was the most recent one?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+    str or None
+        The matched alphabase modification name (e.g. 'Oxidation@M'), or ``None``
+        if no candidate matches the mass and residue/terminus within tolerance.
+
+    """
+    is_terminal = aa_or_term in (
+        ModificationKeys.ANY_N_TERM,
+        ModificationKeys.ANY_C_TERM,
+    )
+    term_keyword = "N-term" if aa_or_term == ModificationKeys.ANY_N_TERM else "C-term"
```
GeorgWa: Please add a constant for "N-term" is this matching a peaks term or a subset of the alphabase N-term?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+        """Load the wide PEAKS export and explode it into long (per-fragment) format."""
+        # Explicit tab separator: the 'Activation Mode' value contains a comma
+        # (e.g. "CID, CAD(y and b ions)") which would fool delimiter sniffing.
+        wide_df = pd.read_csv(filename, sep="\t", keep_default_na=False)
+        return self._explode_peaks_list(wide_df)
+
+    def _explode_peaks_list(self, wide_df: pd.DataFrame) -> pd.DataFrame:
+        """Explode the packed ``Peaks List`` column into one row per fragment."""
+        sequences = wide_df[_PEAKS_SEQUENCE_COLUMN].to_numpy()
+        charges = wide_df["z"].to_numpy()
```
GeorgWa: Please add constant class for peaks column names.

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+        self._psm_df[PsmDfCols.MOD_SITES] = sites_list
+
+        if unknown_mods:
+            warnings.warn(
+                f"Unknown PEAKS modifications: {sorted(unknown_mods)}. Precursors "
+                f"with these modifications will be removed. Add them to "
+                f"'mass_mapped_mods' for 'peaks_library' in psm_reader.yaml to keep them."
+            )
+
+    def _parse_peaks_modifications(
```
GeorgWa: I think it would be a bit easier to structure if the modification translation was it's own class.
This way the column parsing and transformation would be more of a high level transformation with the mod mapping business logic being encapsulated.

GeorgWa (commented):
Looks good!

alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -748,6 +748,38 @@ library_reader_base:
...
+# PEAKS Studio DIA library export (wide format: one row per precursor, all
+# fragments packed into a single "Peaks List" cell). PeaksLibraryReader explodes
+# it into the long format below before the standard library pipeline runs, and converts it to the dense SpeclibBase format.
+
+peaks_library:
+  reader_type: peaks_library
+  rt_unit: second
+  fixed_C57: False
+  column_mapping:
+    'sequence': 'Sequence (backbone)'
```
mschwoer, line 761: please indicate the version of peaks you tested these columns with as a comment

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+        column_mapping: Optional[dict] = None,
+        modification_mapping: Optional[dict] = None,
+        fdr: float = 0.01,
+        fixed_C57: bool = False,  # noqa: N803, FBT001, FBT002
+        mod_seq_columns: Optional[List[str]] = None,
+        rt_unit: Optional[str] = None,
+        precursor_mz_min: float = 400,
+        precursor_mz_max: float = 2000,
+        decoy: Optional[str] = None,
+        **kwargs,
```
mschwoer: do we need to allow kwargs?

alphabase/spectral_library/peaks_reader.py
mschwoer: shouldn't these column names all come from the `yaml`?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+        records = []
+        for sequence, charge, precursor_mz, rt, mod_str, peaks_list in zip(
+            sequences, charges, precursor_mzs, rts, modifications, peaks_lists
+        ):
+            if not peaks_list:
+                continue
+            for peak in peaks_list.split(";"):
+                fields = peak.split(":")
+                if len(fields) != 3:  # noqa: PLR2004 mz:intensity:annotation
+                    continue
```
mschwoer, line 375: should we warn here that an entry is not readable? (same in ln l.227)

Maybe collect all errors and issue one warning at the end

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+                frag_type, frag_number, frag_charge, loss_type = parsed
+                records.append(
+                    (
+                        sequence,
+                        charge,
+                        precursor_mz,
+                        rt,
+                        mod_str,
+                        float(mz_str),
+                        float(intensity_str),
```
mschwoer, line 390: will these float conversions always work?

tests/unit/psm_reader/test_peaks_library_reader.py
```diff
@@ -0,0 +1,241 @@
...
+
+class TestEndToEnd:
+    """Tests for the full import pipeline on a written sample library."""
+
+    def test_modifications_are_parsed(self, sample_library):
+        reader = PeaksLibraryReader()
+        reader.import_file(sample_library)
+
+        precursor_df = reader.precursor_df.set_index("sequence")
+        assert len(precursor_df) == 3
```
mschwoer: please use `pd.testing.assert_frame_equals()` on the whole `df` to make this test stronger (and more readable ;-))

(also below)

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+import numpy as np
+import pandas as pd
+
+from alphabase.constants.modification import MOD_MASS, ModificationKeys
+from alphabase.psm_reader.keys import PsmDfCols
+from alphabase.psm_reader.psm_reader import psm_reader_yaml
+from alphabase.spectral_library.reader import LibraryReaderBase
+
+# Column names in the raw PEAKS export.
+_PEAKS_SEQUENCE_COLUMN = "Sequence (backbone)"
```
mschwoer: shouldn't those fome from the `yaml`?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+_PEAKS_MOD_COLUMN = "Modifications"
+_PEAKS_PEAKS_COLUMN = "Peaks List"
+
+# Internal (synthesized) long-format fragment column names, mapped in psm_reader.yaml.
+_FRAGMENT_MZ = "FragmentMz"
```
mschwoer: shouldn't those fome from the yaml?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,361 @@
...
+        self._mod_mass_tol = reader_config.get("mod_mass_tol", 0.1)
+
+    def _load_file(self, filename: str) -> pd.DataFrame:
+        """Load the wide PEAKS export and explode it into long (per-fragment) format."""
+        # Explicit tab separator: the 'Activation Mode' value contains a comma
+        # (e.g. "CID, CAD(y and b ions)") which would fool delimiter sniffing.
+        wide_df = pd.read_csv(filename, sep="\t", keep_default_na=False)
+        return self._explode_peaks_list(wide_df)
+
+    def _explode_peaks_list(self, wide_df: pd.DataFrame) -> pd.DataFrame:
```
mschwoer, line 323: `@staticmethod`?

alphabase/spectral_library/peaks_reader.py
mo-sameh: I'd been thinking about this too. I tthink there are 3 patterns we have for mod matching:
- Literal matching (as in **MaxQuant**): the token from the search engine is looked up in rev_mod_mapping defined by `modification_mappings`.
- Full-DB mass matching (as in **Sage**): searches the whole `MOD_DF` by mass + residue with a ppm tolerance.
- Limited candidates mass matching(as in **MSFragger**): matches the mass only against a candidate list of alphabase mods declared in the yaml (`mass_mapped_mods`), rather than the full `MOD_DF`.

I thought the third pattern is the safest since `MOD_DF` can have multiple entries with the same mass and residue. Sage just warns and picks the lowest `unimod_id`, but if we keep the candidate list we avoid the ambiguity by choosing the correct modifications in `mass_mapped_mods`. What do you think?

alphabase/spectral_library/peaks_reader.py
mo-sameh: Will read them from the yaml instead

alphabase/spectral_library/peaks_reader.py
mo-sameh: I added them as alphabase level constant

alphabase/spectral_library/peaks_reader.py
mo-sameh: No, removed

alphabase/spectral_library/peaks_reader.py
mo-sameh: Yes absolutely!

alphabase/spectral_library/peaks_reader.py
mo-sameh: now it reads from self.column_mapping, so kept as an instance method.

alphabase/spectral_library/peaks_reader.py
mo-sameh: They should always be numeric. Which is the case in all the files from the Sciex experiment.

alphabase/spectral_library/peaks_reader.py
mo-sameh: I now log upto 5 examples of the non-readable entries, instead of all of them. How does this look?

alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -748,6 +748,42 @@ library_reader_base:
     - 'FullUniModPeptideName'
   modification_mapping_type: 'maxquant'
 
+# PEAKS Studio DIA library export (wide format: one row per precursor, all
+# fragments packed into a single "Peaks List" cell). PeaksLibraryReader explodes
+# it into the long format below before the standard library pipeline runs, and converts it to the dense SpeclibBase format.
+# Columns tested against PEAKS Studio 13.5.
+
+peaks_library:
```
GeorgWa, line 756: Do we now have two different patterns? The peaks reader is a standalone class and config, the DIANN reader is an extension of the LibraryReaderBase.

Whats the best practice in your opinion?

alphabase/constants/modification.py
```diff
@@ -62,6 +62,12 @@ class ModificationKeys(metaclass=_ConstantsClass):
     #: Protein C-terminus only
     PROTEIN_C_TERM: str = "Protein_C-term"
 
+    #: Suffix shared by all N-terminal sites ("Any_N-term", "Protein_N-term")
+    N_TERM: str = "N-term"
```
mschwoer, line 66: shouldn't those constants then be used also in l. 54-63?

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,432 @@
...
+            match = _MOD_ENTRY_PATTERN.match(entry)
+            if match is None:
+                raise ValueError(
+                    f"Invalid PEAKS modification entry '{entry}': expected "
+                    f"'<position><Name>-(<mass>)' (e.g. '9-Oxidation (M)-(15.99)')."
+                )
+            position, name, mass_str = match.groups()
+            mass = float(mass_str)
+
+            if ModificationKeys.N_TERM in name:
```
mschwoer: `name` is PEAKS data, correct? and they happen to use the same naming convention ("N-Term") as alphabase does?
Then we should use a PEAKS-specific constant here (although the value is the same).
Reason: if at one day we choose to change our naming scheme (i.e. the value of `ModificationKeys.N_TERM`), this breaks.

[or maybe I got something wrong here]

alphabase/spectral_library/peaks_reader.py
```diff
@@ -0,0 +1,432 @@
...
+            The raw PEAKS export, one row per precursor.
+
+        Returns
+        -------
+        pd.DataFrame
+            Long-format dataframe with one row per fragment, ready for the standard
+            column translation and library pipeline.
+
+        """
+        cm = self.column_mapping
```
mschwoer: (nit) `cm -> column_mapping` (unless you plan a FORTRAN77 port :-D )

mschwoer (approved):

alphabase/constants/const_files/psm_reader.yaml
mo-sameh: Both use the same pattern. The DIANN reader can directly use the `LibraryReaderBase`. While the peaks reader extends `LibraryReaderBase` and modify the loading to convert the wide (`peak_list`) format to the long format expected.

alphabase/constants/modification.py
mo-sameh: These are a level more general than 54-63 (the bare terminal suffix, not a full site name) the same way ANY_N_TERM and ANY_N_TERM_SPECIFIC = "^Any_N-term" are at different levels.

alphabase/spectral_library/peaks_reader.py
mo-sameh: You are right. I've decoupled it and used a separate constant for peaks. I kept the new alphabase constants since they are used for an alphabase-vs-alphabase check in the Sage.

PR #425, opened 2026-08-17
move hdf out of full extra

mschwoer:
partially revert https://github.com/MannLabs/alphabase/pull/407

### DIFF
--- a/.github/workflows/pip_installation.yml
+++ b/.github/workflows/pip_installation.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest, macos-latest-xlarge]
-        python-version : [3.9]
+        python-version : [3.9, 3.11]
     uses: ./.github/workflows/_run_tests.yml
     with:
       os: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


lucas-diedrich:
Why revert?

mschwoer:
@lucas-diedrich https://github.com/MannLabs/alphabase/pull/407#discussion_r3758058213

.github/workflows/pip_installation.yml
```diff
@@ -29,12 +29,39 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest, macos-latest-xlarge]
+        python-version : [3.9]
```
lucas-diedrich: should we also test on the upper bound (python 3.14?) - as python 3.9 behaves very differently from python 3.11+

lucas-diedrich (approved):

lucas-diedrich:
The commit message (Update numba availability check and add tests for truncate_isotope function) is very different from the code changes - is this intended?

.github/workflows/pip_installation.yml
mschwoer: we can give it a try .. I chose 3.11 now as this is the one we ship AlphaDIA with ..

PR #427, opened 2026-08-21
fix: encode heavy isotopes in TMT6/10plex SMILES annotations

Dondonn:
The smiles column for TMT6plex/TMT10plex encoded the light tag (all 12C/14N), disagreeing with the declared composition H(20)C(8)13C(4)N(1)15N(1)O(2) by ~5 Da. Add [13CH]/[13CH3]/[15N] labels to the dimethylpiperidine reporter so the SMILES matches the composition, and unify the malformed TMT10plex@Any_N-term skeleton with its TMT6plex twin. Verified each mod delta reproduces the declared composition exactly (H included).

---

- [x] I have read and agree to the [MannLabs Contributor License Agreement](https://github.com/MannLabs/.github/blob/main/CLA.md)

### DIFF
--- a/alphabase/constants/const_files/modification.tsv
+++ b/alphabase/constants/const_files/modification.tsv
@@ -1027,19 +1027,19 @@ Ethanolamine@D	43.042199	43.0678	H(5)C(2)N(1)	0.0		Chemical derivative	734		0.0
 Ethanolamine@Any_C-term	43.042199	43.0678	H(5)C(2)N(1)	0.0		Chemical derivative	734		0.0
 Ethanolamine@E	43.042199	43.0678	H(5)C(2)N(1)	0.0		Chemical derivative	734		0.0
 Ethanolamine@C	43.042199	43.0678	H(5)C(2)N(1)	0.0		Chemical derivative	734		0.0
-TMT6plex@T	229.162932	229.2634	H(20)C(8)13C(4)N(1)15N(1)O(2)	0.0		Isotopic label	737	CC1N(CC(NCCC(=O)O[C@H](C)[C@@H](N([Fl])([Fl]))C(=O)[Ts])=O)C(C)CCC1	0.0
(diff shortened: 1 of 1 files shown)


Dondonn:
Added TMTpro in 3995d20.

**Two defects there, not one.** `TMTpro@K` has the same problem as TMT6/10plex: `composition` declares `H(25)C(8)13C(7)N(1)15N(2)O(3)` while the SMILES is all-light, leaving it 9.0176 Da short (7×13C + 2×15N). `TMTpro@Any_N-term` is additionally malformed — it ships as a bare skeleton, `C(=O)CCCNC(=O)CCCNC(=O)C1N(CC(C)C)CCC1`, with no `[Lv]` attachment placeholder unlike every other N-term entry, so the peptide builder bonds the N-terminus to whatever atom the string happens to start with and the delta comes out **+19.0137 Da** rather than merely light. It is replaced with `TMTpro@K`'s skeleton terminated in `[Lv]`, after which labelling reaches the reference mass.

**On isotope placement**, since the table can hold only one SMILES per modification and therefore encodes one channel. All channels of a tag are isobaric, so this affects no mass a normal workflow computes — the intact tag, the precursor and every b/y ion are identical whichever is chosen. It shows up only in a fragment produced by a cut *inside* the tag, and to anyone reading the structure.

- **TMT6plex / TMT10plex** (56f4114): reporter-labelled — all four `13C` and the `15N` on the 2,6-dimethylpiperidine, i.e. channel **131N**.
- **TMTpro** (3995d20): balancer-labelled, i.e. channel **126**. It cannot follow the same convention: TMTpro needs two `15N` and its pyrrolidine reporter contains one nitrogen, so even the most reporter-heavy channel (134N) still strands one `15N` on the balancer. Its balancer holds exactly the 7 C and 2 N that `13C(7) 15N(2)` calls for, which makes 126 the one placement the composition fixes by itself rather than leaving to convention. Happy to switch it to 134N if you would rather the entries read uniformly.

Both TMTpro entries re-encode to `unimod_mass` 304.207146 through `PeptideSmilesEncoder`. `TMTpro@{T,S,H,Protein_N-term}` carry no SMILES at all and are untouched — they would need structures supplied before they can be encoded.

GeorgWa (approved):
Looks good!
Verified and checked the smiles

PR #428, opened 2026-08-31
Share the modification registry with multiprocessing workers

GeorgWa:
## Problem

A worker starts with `spawn` and imports alphabase again. It has only the modifications from `modification.tsv`. Run-time changes are absent: custom modifications, modloss filtering, lower-case AAs, a custom TSV.

`calc_precursor_isotope_intensity` used a fallback to one process. The other multiprocessing functions used the incorrect registry and gave no warning.

## Change

`_spawn_pool()` copies `MOD_DF` in the parent and installs it in each worker, with `get_modification_state()` and `set_modification_state()`. The fallback and `has_custom_mods()` are removed.

The copy is complete, so a new column reaches the workers without a change here. #303 sent selected fields and thus needed a change for each new field.

`parallel_imap()` and `parallel_apply()` in `alphabase/utils.py` are now the only functions that make a pool. Converted: `calc_precursor_isotope_info_mp`, `calc_precursor_isotope_intensity_mp`, `_apply_translate_modifications_mp`, `SpecLibDecoy.translate_to_decoy`. This removes three copies of `_batchify_df` and 125 lines.

Also corrects `calc_precursor_isotope_info`, which sent `process_bar=` to a parameter with the name `progress_bar` and thus raised `TypeError`.

## Tests

`test_worker_registry_matches_parent` compares the registry of the worker with the registry of the parent, after each of four run-time changes. It fails without this PR.

### DIFF
--- a/alphabase/constants/modification.py
+++ b/alphabase/constants/modification.py
@@ -1,3 +1,11 @@
+"""Modification registry.
+
+`MOD_DF` is the single source of truth: every other module-level lookup here is
+derived from it and rebuilt in place by :func:`update_all_by_MOD_DF`. Anything
+that mutates the registry must go through that function, which is what makes
+:func:`get_modification_state` a complete snapshot of the registry.
+"""
+
 import os
 from typing import List, Union
 
@@ -463,7 +471,7 @@ def _check_mass_sanity(
     composition_mass = calc_mass_from_formula(composition)
     if not np.allclose(composition_mass, MOD_MASS[mod_name], atol=1e-5):
         raise ValueError(
-            f"Modification mass of {mod_name} is inconsistent with the composition formula: {composition}, df version {MOD_DF.loc[mod_name,['composition']]}"
+            f"Modification mass of {mod_name} is inconsistent with the composition formula: {composition}, df version {MOD_DF.loc[mod_name, ['composition']]}"
             f" calculated_mass={composition_mass}, mod_mass={MOD_MASS[mod_name]}"
         )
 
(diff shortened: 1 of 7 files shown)


jalew188 (approved):
LGTM

alphabase/constants/modification.py
```diff
@@ -533,6 +540,35 @@ def has_custom_mods():
     return len(MOD_DF[MOD_DF["classification"] == _MOD_CLASSIFICATION_USER_ADDED]) > 0
 
 
+def get_modification_state() -> pd.DataFrame:
```
mschwoer, line 543: maybe make explcit that these methods target multiprocessing ..

alphabase/utils.py
```diff
@@ -70,3 +76,121 @@ def _get_delimiter(file_path: str) -> str:
...
+
+    A worker starts with the "spawn" method and imports alphabase again. Thus it
+    loses all changes that were made to the registry at run time. This function
+    copies the registry one time, then installs it in each worker at start.
+    """
+    ctx = context if context is not None else mp.get_context("spawn")
+    return ctx.Pool(
+        processes,
+        initializer=set_modification_state,
+        initargs=(get_modification_state(),),
```
mschwoer, line 92: do I get correctly that `get_modification_state()` reads the `MOD_DF` from the main threads, and `set_modification_state()` gets it as input arg on the spawned threads?

alphabase/constants/modification.py
```diff
@@ -533,6 +540,35 @@ def has_custom_mods():
...
+    give the process all changes that were made at run time.
+
+    Parameters
+    ----------
+    mod_df : pd.DataFrame
+        A copy from :func:`get_modification_state`.
+    """
+    global MOD_DF
+    MOD_DF = mod_df
+    update_all_by_MOD_DF()
```
mschwoer, line 569: this mutates `MOD_INFO_DICT`, `MOD_CHEM`, etc ..  does this work as intended on the worker threads?

alphabase/utils.py
```diff
@@ -70,3 +76,121 @@ def _get_delimiter(file_path: str) -> str:
...
+def _batchify(obj, batch_size: int, group_by=None):
+    """Divide a DataFrame or Series into batches of rows."""
+    groups = (group for _, group in obj.groupby(group_by)) if group_by else iter((obj,))
+    for group in groups:
+        for i in range(0, len(group), batch_size):
+            yield group.iloc[i : i + batch_size]
+
+
+def _batch_count(obj, batch_size: int, group_by=None) -> int:
+    sizes = obj.groupby(group_by).size().values if group_by else [len(obj)]
```
mschwoer, line 105: here we duplicate logic of `_batchify` .. could this be returned by `_batchify`>

PR #429, opened 2026-08-31
Reduce peak memory of flatten_fragments

GeorgWa:
`flatten_fragments` allocated every flat column at dense length (`n_fragment_rows * n_fragment_types`), assembled a dataframe, then copied the kept subset out of it. Pointer reannotation added an `int64` cumulative sum over all dense slots. `parse_fragment` held the number, the position and the precursor length of every dense slot. For top-k libraries, `mz == 0` padding fills most dense slots.

- Build all columns except `mz` and `intensity` from the kept indices.
- Reannotate the pointers with `np.searchsorted` on the kept indices, instead of a dense cumulative sum.
- Accumulate the masks in place.
- Hold the number, the position and the precursor length per fragment row instead of per dense slot, because every charged fragment type of a row shares them. uint16 holds every value, which `max_frag_per_peptide` bounds.
- Compute the fragment numbers for the kept fragments only, instead of in a dense pass over every slot.
- Pass the intensities without a copy. Zeroing the intensity of padding slots was a no-op, because padding is always excluded.

Peak RSS and runtime, `keep_top_k_fragments=12`:

| dense slots | peak main | peak PR | time main | time PR |
|---|---|---|---|---|
| 120M | 6.45 GB | 1.76 GB | 7 s | 3 s |
| 480M | 23.22 GB | 6.08 GB | 15 s | 7 s |
| 1.48G | 58.73 GB | 16.08 GB | 45 s | 14 s |

### DIFF
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -1040,14 +1040,10 @@ def flatten_fragments(
 
     n_fragment_types = len(fragment_mz_df.columns)
 
-    # Only `mz` and `intensity` are required at the full dense length of
-    # n_fragment_rows * n_fragment_types, as they determine which fragments are
-    # retained. Every other flat column is built directly from the indices of the
-    # retained fragments, so neither the dense intermediates for
-    # type/loss_type/charge/number/position nor the copy that filtering an
-    # assembled dataframe would make are ever materialised. Peak memory is thus
-    # driven by the number of retained fragments instead of the number of dense
-    # slots, which for top-k libraries is dominated by mz==0 padding.
+    # Only mz and intensity need the full dense length, because they give the
+    # keep mask. Dense arrays for the other columns and a copy from a filtered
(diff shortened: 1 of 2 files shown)


alphabase/peptide/fragment.py
```diff
@@ -1037,99 +1251,64 @@ def flatten_fragments(
...
-    if "charge" in custom_columns:
-        frag_df["charge"] = np.array(
-            np.tile(frag_charges, len(fragment_mz_df)), dtype=np.int8
-        )
-
-    frag_directions = np.array(
-        np.tile(frag_directions, (len(fragment_mz_df), 1)), dtype=np.int8
+        if use_intensity
+        else None
     )
```
Desperadus, line 1273: You say: "Nothing writes to `mz` or `intensity`" - should we enforce it like:

```python
mz.flags.writeable = False
if intensity is not None:
    intensity.flags.writeable = False
```
?

Desperadus (approved):
Awesome mem improvement - LGTM

PR #434, opened 2026-08-31
[1/5] Tests: characterize flatten_fragments before the memory refactor

GeorgWa:
**Tests only. No production change.**

`flatten_fragments` had no unit test. Add black-box tests that pin its contract, so the refactoring that follows can be checked step by step.

All tests pass unchanged on `main`.

First of 5 stacked PRs splitting #429.

### DIFF
--- a/tests/unit/peptide/test_fragment.py
+++ b/tests/unit/peptide/test_fragment.py
@@ -1,9 +1,15 @@
 from unittest.mock import patch
 
+import numpy as np
+import pandas as pd
 import pytest
 
 from alphabase.peptide.fragment import (
+    FRAGMENT_TYPES,
+    PEAK_INTENSITY_DTYPE,
+    PEAK_MZ_DTYPE,
     filter_valid_charged_frag_types,
+    flatten_fragments,
     parse_charged_frag_type,
 )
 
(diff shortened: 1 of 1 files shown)


tests/unit/peptide/test_fragment.py
```diff
@@ -51,3 +57,271 @@ def test_filter_valid_charged_frag_types(mock_parse):
...
+    )
+
+    assert len(precursor_df) == 0
+    assert len(frag_df) == 0
+
+
+@pytest.mark.requires_numba
+def test_flatten_fragments_long_precursor():
+    """A long precursor keeps fragment numbers above 255, and the columns stay uint32."""
+    n_rows = 260  # close to the max_frag_per_peptide limit of _fill_in_indices
```
mschwoer, line 307: brittle?

tests/unit/peptide/test_fragment.py
```diff
@@ -51,3 +57,271 @@ def test_filter_valid_charged_frag_types(mock_parse):
...
+        precursor_df,
+        pd.DataFrame(mz, columns=CHARGED_FRAG_TYPES),
+        pd.DataFrame(intensity, columns=CHARGED_FRAG_TYPES),
+    )
+
+
+def _expected_keep_mask(
+    precursor_df, mz_df, intensity_df, keep_top_k_fragments, min_fragment_intensity
+):
+    """Give the keep mask over all dense slots. This does not use flatten_fragments."""
```
mschwoer, line 106: could we add a test for this?

mschwoer: seriously, could we not work with statically define input data (and data to assert against)?

such minimal data models would also help understand/debug the code better one day

GeorgWa: @mschwoer do you mean we SHOULD work with static examples or we should keep this logic?

mschwoer: we should "strongly consider"™️ it :-D

tests/unit/peptide/test_fragment.py
GeorgWa: fixed it

tests/unit/peptide/test_fragment.py
GeorgWa: changedf it to use a static fixture

tests/unit/peptide/test_fragment.py
```diff
@@ -51,3 +57,415 @@ def test_filter_valid_charged_frag_types(mock_parse):
...
+    """flatten_fragments annotates each kept slot with its dense position."""
+    # Given
+    precursor_df, mz_df, intensity_df = minimal_library
+
+    # When
+    _, frag_df = flatten_fragments(precursor_df, mz_df, intensity_df)
+
+    # Then
+    # kept slots: p0 r0 b, p0 r0 y, p0 r1 y, p1 r0 b, p1 r0 y, p1 r1 b
+    np.testing.assert_array_equal(
```
mschwoer: why not assert_frame_equal?

mschwoer (approved):

PR #435, opened 2026-08-31
[2/5] Refactor: make the fragment parsing helpers private

GeorgWa:
**Pure refactor. Rename only, no functional change.**

`parse_fragment`, `fill_in_indices` and `calculate_fragment_numbers` only serve `flatten_fragments`, and nothing in alphabase, alphadia or peptdeep calls them. A leading underscore states that, and frees their signatures for the refactoring that follows.

Stacked on #434. Splitting #429.

### DIFF
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -802,7 +802,7 @@ def mask_fragments_for_charge_greater_than_precursor_charge(
 
 
 @numba_njit(parallel=True)
-def fill_in_indices(
+def _fill_in_indices(
     frag_start_idxes: np.ndarray,
     frag_stop_idxes: np.ndarray,
     indices: np.ndarray,
@@ -877,7 +877,7 @@ def fill_in_indices(
 @numba_vectorize(
     [nb_.uint32(nb_.int8, nb_.uint32, nb_.uint32, nb_.uint32)], target="parallel"
 )
-def calculate_fragment_numbers(
+def _calculate_fragment_numbers(
     frag_direction: np.int8,
     frag_number: np.uint32,
     index: np.uint32,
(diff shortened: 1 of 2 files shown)


mschwoer (commented):
LGTM

alphabase/peptide/fragment.py
mschwoer, line 919: "fragment numbers" is kinda ambigous.. is this `counts`? or `enumeration`? not sure if it should be renamed now (depends on the radius of this change), but at last a comment would help :-)

mo-sameh:
PS: I'll be creating a PR soon to optimize the translation modules (`translate.py`, `translate_diann.py`). They each have their own fragment-flattening code (merge_precursor_fragment_df) that does very similar task as flatten_fragments, just emitting the type/loss/charge columns as strings for DIA-NN/Spectronaut rather than numeric codes. 

I think it will be better to reuse the much more optimized helpers here instead.
So maybe worth keeping them public? Not a blocker, just flagging since there'll be a caller shortly.

GeorgWa:
> PS: I'll be creating a PR soon to optimize the translation modules (`translate.py`, `translate_diann.py`). They each have their own fragment-flattening code (merge_precursor_fragment_df) that does very similar task as flatten_fragments, just emitting the type/loss/charge columns as strings for DIA-NN/Spectronaut rather than numeric codes.
> 
> I think it will be better to reuse the much more optimized helpers here instead. So maybe worth keeping them public? Not a blocker, just flagging since there'll be a caller shortly.

very nice!

alphabase/peptide/fragment.py
GeorgWa: I agree, it's the MS convention fragment numbering. I will add a comment :D

GeorgWa: I agree, it's the MS convention fragment numbering. I will add a comment :D

mschwoer (approved):

PR #436, opened 2026-08-31
[3/5] Refactor: extract the charged fragment type annotation and the pointer reannotation

GeorgWa:
**Pure refactor. No functional change, output bit-identical to main.**

Move two concerns out of the body of `flatten_fragments`, so the memory changes that follow are diffs in named helpers instead of a long function body.

- Return the series id, loss id, charge and direction as typed arrays instead of four parallel lists of Python ints. Tiling a typed array keeps its dtype, so the dense type/loss_type/charge/direction arrays no longer pass through an `int64` intermediate of 8 bytes per dense slot.

Stacked on #435. Splitting #429.

### DIFF
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -802,7 +802,7 @@ def mask_fragments_for_charge_greater_than_precursor_charge(
 
 
 @numba_njit(parallel=True)
-def fill_in_indices(
+def _fill_in_indices(
     frag_start_idxes: np.ndarray,
     frag_stop_idxes: np.ndarray,
     indices: np.ndarray,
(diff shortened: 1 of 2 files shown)


alphabase/peptide/fragment.py
```diff
@@ -1052,37 +1122,23 @@ def flatten_fragments(
...
-        frag_type, charge = parse_charged_frag_type(charged_frag_type)
-        frag_charges.append(charge)
-        frag_types.append(FRAGMENT_TYPES[frag_type].series_id)
-        frag_loss_types.append(FRAGMENT_TYPES[frag_type].loss_id)
-        frag_directions.append(FRAGMENT_TYPES[frag_type].direction_id)
+    series_ids, loss_ids, charges, directions = _annotate_charged_frag_types(
+        fragment_mz_df.columns.values
+    )
 
+    # tiling the typed arrays keeps their dtype. A tiled list of Python ints gave
```
mschwoer, line 1129: "tiling the typed arrays keeps their dtype" -> I feel this is redundant? like "`a = 1+1 # this will stay an int`" ? also, second part of comment references previous implementation.

(I am not saying the information is not valuable, it's just that code is not the right place for it .. this would be a good example of the PR author commenting his own code to guide the review)

mschwoer (approved):

PR #437, opened 2026-08-31
[4/5] Perf: build the flat fragment columns from the kept indices

GeorgWa:
**Performance. Output bit-identical to main.**

`flatten_fragments` allocated every flat column at dense length (`n_fragment_rows * n_fragment_types`), assembled a dataframe, then copied the kept subset out of it. Pointer reannotation added an `int64` cumulative sum over all dense slots. For top-k libraries, `mz == 0` padding fills most dense slots.

- Build all columns except `mz` and `intensity` from the kept indices.
- Reannotate the pointers with `np.searchsorted` on the kept indices, instead of a dense cumulative sum.
- Accumulate the masks in place.

48M dense slots: peak RSS 3.49 GB -> 1.77 GB.

Stacked on #436. Splitting #429.

### DIFF
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -802,7 +802,7 @@ def mask_fragments_for_charge_greater_than_precursor_charge(
 
 
 @numba_njit(parallel=True)
-def fill_in_indices(
+def _fill_in_indices(
     frag_start_idxes: np.ndarray,
     frag_stop_idxes: np.ndarray,
     indices: np.ndarray,
@@ -877,7 +877,7 @@ def fill_in_indices(
 @numba_vectorize(
     [nb_.uint32(nb_.int8, nb_.uint32, nb_.uint32, nb_.uint32)], target="parallel"
 )
-def calculate_fragment_numbers(
+def _calculate_fragment_numbers(
     frag_direction: np.int8,
     frag_number: np.uint32,
     index: np.uint32,
@@ -907,7 +907,7 @@ def calculate_fragment_numbers(
     return frag_number
 
 
-def parse_fragment(
+def _parse_fragment(
     frag_directions: np.ndarray,
     frag_start_idxes: np.ndarray,
     frag_stop_idxes: np.ndarray,
@@ -954,7 +954,7 @@ def parse_fragment(
     )
 
     # Fill in indices, max indices and excluded indices
-    fill_in_indices(
+    _fill_in_indices(
         frag_start_idxes,
         frag_stop_idxes,
         indices,
@@ -966,7 +966,7 @@ def parse_fragment(
     )
 
     # Calculate fragment numbers
-    frag_numbers = calculate_fragment_numbers(
+    frag_numbers = _calculate_fragment_numbers(
         frag_directions, frag_numbers, indices, max_indices
     )
     return frag_numbers, indices, excluded_indices
@@ -1081,7 +1081,7 @@ def flatten_fragments(
         np.tile(frag_directions, (len(fragment_mz_df), 1)), dtype=np.int8
     )
 
-    numbers, positions, excluded_indices = parse_fragment(
+    numbers, positions, excluded_indices = _parse_fragment(
         frag_directions,
         precursor_df.frag_start_idx.values,
         precursor_df.frag_stop_idx.values,
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -1008,38 +1008,180 @@ def _annotate_charged_frag_types(
     )
 
 
-def _reannotate_precursor_pointers(
-    precursor_df: pd.DataFrame, excluded: np.ndarray, n_fragment_types: int
-) -> None:
-    """Add the flat fragment pointers to `precursor_df`, in place.
+def _select_dense_fragments(
+    precursor_df: pd.DataFrame,
+    mz: np.ndarray,
+    intensity: Union[np.ndarray, None],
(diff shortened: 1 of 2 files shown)


alphabase/peptide/fragment.py
```diff
@@ -1107,69 +1249,59 @@ def flatten_fragments(
...
+        min_fragment_intensity,
     )
 
-    if "number" in custom_columns:
-        frag_df["number"] = numbers.reshape(-1)
-
-    if "position" in custom_columns:
-        frag_df["position"] = positions.reshape(-1)
-
+    frag_df = {}
```
mschwoer, line 1282: `frag_dict`?

alphabase/peptide/fragment.py
```diff
@@ -1008,38 +1008,180 @@ def _annotate_charged_frag_types(
...
+
+    numbers, positions, not_top_k = _parse_fragment(
+        dense_directions,
+        precursor_df.frag_start_idx.values,
+        precursor_df.frag_stop_idx.values,
+        keep_top_k_fragments,
+        intensity,
+        n_fragment_types,
+    )
+    del dense_directions
```
mschwoer, line 1071: did you check if this is garbage collected immediately? or does this need an explicit call to `gc`?

alphabase/peptide/fragment.py
```diff
@@ -1008,38 +1008,180 @@ def _annotate_charged_frag_types(
...
+
+    if intensity is None:
+        excluded = mz == 0
```
mschwoer, line 1074: `mz == 0` is shorthand for `all array elements are zero`? is there a more numpythonic way of doing that?

alphabase/peptide/fragment.py
```diff
@@ -1008,38 +1008,180 @@ def _annotate_charged_frag_types(
...
+    else:
+        is_padding = mz == 0
+        intensity[is_padding] = 0.0
+        # in-place operations prevent more dense boolean arrays
+        excluded = intensity < min_fragment_intensity
+        excluded |= is_padding
+        excluded |= not_top_k
+        del is_padding
+    del not_top_k
```
mschwoer, line 1083: how large can the `mz` array be? is it worth the `del` noise?

alphabase/peptide/fragment.py
```diff
@@ -1008,38 +1008,180 @@ def _annotate_charged_frag_types(
...
-    cum_sum_tresh = np.zeros(shape=len(excluded) + 1, dtype=np.int64)
-    cum_sum_tresh[1:] = np.cumsum(excluded)
-
-    precursor_df["flat_frag_start_idx"] -= cum_sum_tresh[
-        precursor_df.flat_frag_start_idx.values
-    ]
-    precursor_df["flat_frag_stop_idx"] -= cum_sum_tresh[
-        precursor_df.flat_frag_stop_idx.values
-    ]
+    dense_start_idx = precursor_df.frag_start_idx.values.astype(np.int64)
```
mschwoer, line 1177: please use `[]` notation to access df columns consistently

alphabase/peptide/fragment.py
```diff
@@ -1008,38 +1008,180 @@ def _annotate_charged_frag_types(
...
+    dense_stop_idx = precursor_df.frag_stop_idx.values.astype(np.int64)
```
mschwoer, line 1178: so `frag_stop_idx` is `int32` (I assume), but needs to be `int64` for the dense representation? worth a comment..

alphabase/peptide/fragment.py
GeorgWa: I only checked if this improved memory, not if it triggered gc. but assume so.

alphabase/peptide/fragment.py
GeorgWa: Yes exactly, this is the pythonic way tbh

alphabase/peptide/fragment.py
GeorgWa: yes, and we want to be explicit with the dtype before constructing the pandas dataframes. Otherwise weird parsing can happen.

alphabase/peptide/fragment.py
GeorgWa: This is a bit tricky. we use this pattern a lot as this is how we create a pandas df. calling it frag dict would suggest somethign else.

alphabase/peptide/fragment.py
GeorgWa: While I see the point. I'm not sure if it's worth the 30min investigation.

GeorgWa: removed them

alphabase/peptide/fragment.py
mschwoer: well, `_df` suggests it's a dataframe .. but if this follows current code convention, fair enough

alphabase/peptide/fragment.py
mschwoer: isn't gc only triggered automatically at the exit of a method?

alphabase/peptide/fragment.py
mschwoer: Thank you.

Note: What would take you 30 minutes of investigation now, will take other people (including future you) 90 minutes, nice "every LoC is a liability" example :-)

alphabase/peptide/fragment.py
GeorgWa: I removed the del statements 👍🏻

mschwoer (approved):

PR #438, opened 2026-08-31
[5/5] Perf: index the fragment rows instead of the dense slots

GeorgWa:
**Performance. One documented behaviour change, otherwise bit-identical to main.**

`parse_fragment` held the number, the position and the precursor length of every dense slot.

- Hold the number, the position and the precursor length per fragment row instead of per dense slot, because every charged fragment type of a row shares them. uint16 holds every value, which `max_frag_per_peptide` bounds.
- Compute the fragment numbers for the kept fragments only, instead of in a dense pass over every slot.
- Pass the intensities without a copy. Zeroing the intensity of padding slots was a no-op, because padding is always excluded.

Behaviour change: fragment rows that no precursor covers now report number 0 and position 0. `main` leaves those slots unwritten (`np.empty`), so their values were unspecified.

Peak RSS and runtime of the whole stack, `keep_top_k_fragments=12`:

| dense slots | peak main | peak stack | time main | time stack |
|---|---|---|---|---|
| 120M | 6.45 GB | 1.76 GB | 7 s | 3 s |
| 480M | 23.22 GB | 6.08 GB | 15 s | 7 s |
| 1.48G | 58.73 GB | 16.08 GB | 45 s | 14 s |

Last of 5. Stacked on #437, and lands the same end state as #429.

### DIFF
--- a/alphabase/peptide/fragment.py
+++ b/alphabase/peptide/fragment.py
@@ -805,8 +805,8 @@ def mask_fragments_for_charge_greater_than_precursor_charge(
 def _fill_in_indices(
     frag_start_idxes: np.ndarray,
     frag_stop_idxes: np.ndarray,
-    indices: np.ndarray,
-    max_indices: np.ndarray,
+    row_positions: np.ndarray,
+    row_counts: np.ndarray,
     excluded_indices: np.ndarray,
     top_k: int,
     flattened_intensity: np.ndarray,
@@ -814,9 +814,9 @@ def _fill_in_indices(
     max_frag_per_peptide: int = 300,
 ) -> None:
     """
-    Fill in indices, max indices and excluded indices for each peptide.
-    indices: index of fragment per peptide (from 0 to max_index-1)
-    max_indices: max index of fragments per peptide (number of fragments per peptide)
+    Fill in row positions, row counts and excluded indices for each peptide.
+    row_positions: index of the fragment row within its peptide (from 0 to row count - 1)
+    row_counts: number of fragment rows of the peptide
     excluded_indices: not top k excluded indices per peptide
 
     Parameters
@@ -827,11 +827,13 @@ def _fill_in_indices(
     frag_stop_idxes : np.ndarray
         stop indices of fragments for each peptide
 
-    indices : np.ndarray
-        index of fragment per peptide (from 0 to max_index-1) it will be filled in this function
+    row_positions : np.ndarray
+        index of the fragment row within its peptide, one value per fragment row.
+        It will be filled in this function.
 
-    max_indices : np.ndarray
-        max index of fragments per peptide (number of fragments per peptide) it will be filled in this function
+    row_counts : np.ndarray
+        number of fragment rows of the peptide, one value per fragment row.
+        It will be filled in this function.
 
     excluded_indices : np.ndarray
         not top k excluded indices per peptide it will be filled in this function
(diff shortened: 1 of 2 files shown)


alphabase/peptide/fragment.py
```diff
@@ -936,40 +938,40 @@ def _parse_fragment(
...
     """
-    # Allocate memory for fragment numbers, indices, max indices and excluded indices
-    frag_numbers = np.empty_like(frag_directions, dtype=np.uint32)
-    indices = np.empty_like(frag_directions, dtype=np.uint32)
-    max_indices = np.empty_like(frag_directions, dtype=np.uint32)
+    # uint16 holds every value, because `max_frag_per_peptide` of
+    # `_fill_in_indices` bounds the row positions and the row counts. Rows that no
+    # peptide covers keep a zero.
+    row_positions = np.zeros(n_fragment_rows, dtype=np.uint16)
+    row_counts = np.zeros(n_fragment_rows, dtype=np.uint16)
```
mschwoer, line 958: we intantiate here, then pass to `_fill_in_indices` .. why not instantiate there and return?

alphabase/peptide/fragment.py
```diff
@@ -874,56 +875,57 @@ def _fill_in_indices(
...
 ):
     """
-    Calculate fragment numbers for each fragment based on the fragment direction.
+    Calculate the number of a fragment in its ion series.
 
     Parameters
     ----------
     frag_direction : np.int8
-        directions of fragments for each peptide
+        direction of the fragment type. 'abc' ions count from the first amino
```
mschwoer, line 890: could there be string constants for 'abc' and 'xyz' ?

alphabase/peptide/fragment.py
```diff
@@ -874,56 +875,57 @@ def _fill_in_indices(
...
     """
     if frag_direction == 1:
-        frag_number = index + 1
-    elif frag_direction == -1:
-        frag_number = max_index - index
-    return frag_number
+        return row_position + 1
+    if frag_direction == -1:
+        return row_count - row_position
+    return 0
```
mschwoer, line 909: please add this behaviour to the docstring

alphabase/peptide/fragment.py
```diff
@@ -1135,8 +1130,12 @@ def _annotate_kept_fragments(
         the requested columns, in the column order of the flat fragment dataframe
 
     """
-    needs_frag_type = bool({"type", "loss_type", "charge"}.intersection(custom_columns))
+    needs_frag_type = bool(
+        {"type", "loss_type", "charge", "number"}.intersection(custom_columns)
+    )
+    needs_row = bool({"number", "position"}.intersection(custom_columns))
     kept_frag_types = kept_indices % n_fragment_types if needs_frag_type else None
+    kept_rows = kept_indices // n_fragment_types if needs_row else None
```
mschwoer, line 1138: please double check if a `+1` is required due to flooring

alphabase/peptide/fragment.py
```diff
@@ -1146,9 +1145,12 @@ def _annotate_kept_fragments(
     if "charge" in custom_columns:
         columns["charge"] = charges[kept_frag_types]
     if "number" in custom_columns:
-        columns["number"] = numbers.reshape(-1)[kept_indices]
+        columns["number"] = _calculate_fragment_numbers(
+            directions[kept_frag_types], row_positions[kept_rows], row_counts[kept_rows]
+        )
     if "position" in custom_columns:
-        columns["position"] = positions.reshape(-1)[kept_indices]
+        # the flat column keeps its uint32 dtype
```
mschwoer, line 1152: what extra information (compared to the next line) does this comment carry?

alphabase/peptide/fragment.py
```diff
@@ -1258,8 +1260,13 @@ def flatten_fragments(
     # fills most dense slots.
     mz = fragment_mz_df.values.reshape(-1)
     use_intensity = len(fragment_intensity_df) > 0
+    # `copy=False` keeps a dense copy out of memory whenever the dtype already
```
mschwoer, line 1263: "`copy=False` keeps a dense copy out of memory " .. wording is confusing (I guess aircraft operators were not trained in python, but maybe COBOL?)

alphabase/peptide/fragment.py
GeorgWa: I think this is beyiond the scope of the PR

alphabase/peptide/fragment.py
GeorgWa: implemented it, nicer!

mschwoer (approved):

alphabase/peptide/fragment.py
GeorgWa: removed it

PR #439, opened 2026-08-31
[chore] Base ruff on python 3.9

lucas-diedrich:
Python 3.9 breaks ruff linting. Explicitly set ruff linting to python version 3.9


mschwoer:
supersede by #442

PR #440, opened 2026-08-31
[chore] Ruff checks

lucas-diedrich:
Fix unsafe changes


alphabase/smiles/peptide.py
```diff
@@ -88,7 +88,7 @@ def _get_terminal_mods(self, mods: dict) -> tuple:
             (n_term_mod, c_term_mod)
         """
         n_term_mod = mods.get(0)
-        c_term_mod = mods.get(-1, None)
+        c_term_mod = mods.get(-1)
```
lucas-diedrich, line 91: `.get` defaults to None

lucas-diedrich:
@mschwoer I linted all modules with the same rule set, but `ruff-lint-psm-readers.toml` seems to apply different rules on different modules.

Is this ok as is?

PR #442, opened 2026-08-31
Abandon py 38

mschwoer:
Drop support for python 3.8, requires reformatting and relinting. Done in pair with @lucas-diedrich 

supersedes https://github.com/MannLabs/alphabase/pull/439 and https://github.com/MannLabs/alphabase/pull/431


lucas-diedrich (approved):
Checked all diffs manually.

PR #443, opened 2026-09-01
fix xxhash 4 string hashing, skip numpy-2-incompatible pytables tests

mschwoer:
Fixes issues with the full test suite introduced recently.


tests/integration/test_pg_readers.py
```diff
@@ -108,6 +111,9 @@ def test_import_csv_file(
             PGCols.DECOY_INDICATOR,
         ]
 
+    @pytest.mark.skip(
```
lucas-diedrich: Skip for python v3.9 but not v3.11?

lucas-diedrich (approved):
Thanks for fixing this!

PR #444, opened 2026-09-02
[translate 1/3] Characterize the spectral library exports

mo-sameh:
### Goal

Pin what the SWATH/Spectronaut and DIA-NN exports do today, so the two PRs stacked on top can be shown to change only what they claim to.

### What's here

Tests only — no source changes. A shared fixture builds a library covering N-terminal, C-terminal, internal and loss-bearing modifications, at charges 1–2, so it has both real fragments and empty (m/z 0) slots. On top of that: column names and order, modified-sequence rendering, RT column precedence, the m/z / top-k / `min_frag_nAA` filters, `Flags`, and tsv and parquet round trips.

The 'tsv translate' functionality didnt have unit tests before. 

Some of these pin behaviour that is wrong. Those are marked `CHARACTERIZATION (bug)` in their docstrings and say which later commit changes them — a test that flips in PR 3 is meant to, and the diff there is the evidence.

### DIFF
--- a/tests/unit/spectral_library/test_translate.py
+++ b/tests/unit/spectral_library/test_translate.py
@@ -1,0 +1,510 @@
+"""Characterization tests for the SWATH/Spectronaut transition-list export.
+
+These tests pin the behaviour of `alphabase.spectral_library.translate` as it is
+today, so that the upcoming restructuring can be shown to change nothing.
+
+Six tests deliberately pin *buggy* behaviour that a later commit fixes. Each
+carries a `CHARACTERIZATION (bug)` note in its docstring and is expected to be
+rewritten there:
+
+1. the export zeroes intensities in the caller's library and adds `precursor_mz`
+2. so a second export of the same library silently differs from a fresh one
+3. with the m/z window disabled, empty fragment slots leak into the output
+4. `translate_to_tsv` masks by m/z unconditionally, so disabling the window
+   yields a file whose every fragment is at m/z 0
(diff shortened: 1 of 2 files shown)


tests/unit/spectral_library/test_translate.py
```diff
@@ -0,0 +1,510 @@
...
+1. the export zeroes intensities in the caller's library and adds `precursor_mz`
+2. so a second export of the same library silently differs from a fresh one
+3. with the m/z window disabled, empty fragment slots leak into the output
+4. `translate_to_tsv` masks by m/z unconditionally, so disabling the window
+   yields a file whose every fragment is at m/z 0
+5. `rt_norm_pred` is not accepted as a retention time column
+6. the exploded fragment columns are object dtype, with a string `FragmentCharge`
+
+One more, `test_speclib_to_swath_df_returns_none`, pins a function that a later
+commit removes rather than fixes.
```
mschwoer, line 19: these read a bit weird .. also, remove all refenrences to "later commits"

mschwoer (approved):

tests/unit/spectral_library/test_translate.py
mo-sameh: Yeah my bad, I added these to make the stack easier to review. They do get rewritten by 2/3 and 3/3 as the bugs get fixed, so this intermediate state won't survive the merge, but I see now how weird it reads on its own.

PR #445, opened 2026-09-02
[translate 2/3] Split the shared export machinery out of translate.py

mo-sameh:
### Goal

`translate_diann.py` imported five helpers from `translate.py`, which made the SWATH format the de facto shared library for the DIA-NN one. Move what both formats share into a new `translate_core.py` so neither depends on the other's dialect.

### What's here

`translate_core.py` holds the modified-sequence rendering, the candidate precursor columns, and the fragment flattening. Each export keeps only its own column names and writer, and renames the canonical fragment columns into its own dialect through a dict.

The flattening is where the real change is. It used to filter by *zeroing intensities in the caller's library* and letting a later `min_frag_intensity` drop the rows. One design choice, four symptoms: a default export silently zeroed 60% of the library's intensities, a second export at a wider window returned different fragments than a fresh one, the export wrote a `precursor_mz` column into the caller's frame, and disabling the m/z window exported empty fragment slots at m/z 0. Fragments are now selected on a per-precursor copy and dropped rather than zeroed, which fixes all four.

### Verification

Every export was run against `main`'s modules over 19 configurations and compared including dtypes and index. The first four commits are byte-identical to `main` on all 19, with no test edits. The fifth differs on exactly the four cases that disable the m/z window — the fix — and the characterization tests that pinned the old behaviour change with it.

`pytest` warnings drop from 41 to 7; the `SettingWithCopyWarning` storm was the export writing into batch slices of the caller's library.

### DIFF
--- a/alphabase/spectral_library/translate.py
+++ b/alphabase/spectral_library/translate.py
@@ -1,13 +1,44 @@
+"""Translate AlphaBase spectral libraries to a SWATH/Spectronaut transition list.
+
+The shared export machinery lives in :mod:`alphabase.spectral_library.translate_core`;
+this module holds the SWATH column names, the precursor mapping and the tsv writer. The
+names it used to define are re-exported below, so importing them from here keeps
+working.
+"""
+
 import multiprocessing as mp
 
-import numpy as np
 import pandas as pd
 import tqdm
 
-from alphabase.constants.modification import MOD_DF, ModificationKeys
-from alphabase.numba_wrapper import numba_njit
 from alphabase.spectral_library.base import SpecLibBase
-from alphabase.utils import explode_multiple_columns
+from alphabase.spectral_library.translate_core import (
+    CCS_COLUMNS,
+    MOBILITY_COLUMNS,
+    RT_COLUMNS,
+    create_modified_sequence,
+    first_present_column,
+    is_nterm_frag,
+    mask_fragment_intensity_by_frag_nAA,
+    mask_fragment_intensity_by_mz_,
+    merge_precursor_fragment_df,
+    mod_to_unimod_dict,
+)
+
+__all__ = [
+    # re-exported from translate_core, where these now live
+    "create_modified_sequence",
+    "is_nterm_frag",
+    "mask_fragment_intensity_by_frag_nAA",
+    "mask_fragment_intensity_by_mz_",
+    "merge_precursor_fragment_df",
+    "mod_to_unimod_dict",
+    # this module's own
+    "WritingProcess",
+    "speclib_to_single_df",
+    "speclib_to_swath_df",
+    "translate_to_tsv",
+]
 
 # pandas renamed `to_csv`'s newline argument in 1.5; alphabase does not pin a minimum
 _CSV_NEWLINE = (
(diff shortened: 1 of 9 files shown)


alphabase/spectral_library/translate.py
```diff
@@ -1,235 +1,136 @@
...
+    "mod_to_unimod_dict",
+    "get_precursor_mz",
+    # this module's own
+    "WritingProcess",
+    "speclib_to_single_df",
+    "speclib_to_swath_df",
+    "translate_to_tsv",
+]
+
+# pandas renamed `to_csv`'s newline argument in 1.5; alphabase does not pin a minimum
```
mschwoer: could we not just pin a minimum?

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
+"""Machinery shared by the spectral library export formats.
+
+:module:`alphabase.spectral_library.translate` writes a SWATH/Spectronaut transition list
+and :module:`alphabase.spectral_library.translate_diann` a DIA-NN 1.9.1+ parquet library.
+Both flatten the same alphabase library into one row per precursor/fragment pair and
+differ only in dialect, so the modified-sequence rendering, the fragment selection and
+the candidate precursor columns live here rather than in either format.
+
+Before this module, they lived in ``translate.py``, which made the SWATH format the de
+facto shared library: ``translate_diann`` imported five helpers from it.
```
mschwoer, line 10: please remove references to older code

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
...
+        The first candidate column present, else `default`.
+
+    """
+    for column in candidates:
+        if column in precursor_df.columns:
+            return precursor_df[column]
+    return default
+
+
+# @numba.njit #(cannot use numba for pd.Series)
```
mschwoer, line 116: which argument is the series?

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
...
+    min_frag_nAA: int = 0,  # noqa: N803
+    verbose: bool = True,
+) -> pd.DataFrame:
+    """Flatten each precursor's most intense fragments into one row per fragment.
+
+    Filtering, normalization and selection all happen on a per-precursor copy, so the
+    library's fragment frames are left exactly as they were. Fragments outside the m/z
+    window are dropped rather than zeroed, as are empty fragment slots -- a `*_modloss`
+    column of a precursor whose modification has no loss carries m/z 0, and selecting it
+    would export a fragment that does not exist. An unbounded window is expressed by the
```
mschwoer, line 236: Please rephrase the last sentence

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
...
+        )
+        max_frag_mz = np.inf
+
+    frag_types = []
+    frag_losses = []
+    frag_charges = []
+    frag_masses = []
+    frag_intensities = []
+    frag_numbers = []
+    iters = zip(frag_start_idx, frag_stop_idx)
```
mschwoer, line 298: `frag_start_stop_idx_iterator`?

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
...
+    if verbose:
+        iters = tqdm.tqdm(iters)
+    for start, end in iters:
+        masses = fragment_mz_df.iloc[start:end, :].to_numpy()
+        keep = (masses > 0) & (masses >= min_frag_mz) & (masses <= max_frag_mz)
```
mschwoer, line 303: `keep_mask`

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
...
+    from the end, so row `r` of a precursor with `frag_len` fragment rows is `r + 1`
+    for a b-ion and `frag_len - r` for a y-ion.
+    """
+    return [
+        row + 1 if is_nterm_frag(column) else frag_len - row
+        for row, column in zip(rows, columns)
+    ]
+
+
+def fragment_table(  # noqa: PLR0913
```
mschwoer, line 218: `get_fragment_table` ?

alphabase/spectral_library/translate_core.py
```diff
@@ -0,0 +1,383 @@
...
+        infos = [_get_frag_info_from_column_name(column) for column in columns]
+        types, losses, charges = zip(*infos) if infos else ((), (), ())
+        frag_types.append(types)
+        frag_losses.append(losses)
+        frag_charges.append(charges)
+        frag_masses.append(masses[idx_in_df])
+        frag_intensities.append(intens[idx_in_df])
+        frag_numbers.append(_get_frag_num(columns, rows, frag_len))
+
+    table = pd.DataFrame(
```
mschwoer, line 337: `fragments_df`?

alphabase/spectral_library/translate_core.py
mschwoer: in general, docstrings are very LLM-ish, might be worth a human overhaul :-)

alphabase/spectral_library/translate_core.py
mo-sameh: removed this comment, it was inherited by the old code

alphabase/spectral_library/translate.py
```diff
@@ -1,235 +1,136 @@
+"""Translate AlphaBase spectral libraries to a SWATH/Spectronaut transition list.
+
+The shared export machinery lives in :mod:`alphabase.spectral_library.translate_core`;
+this module holds the SWATH column names, the precursor mapping and the tsv writer. The
+names it used to define are re-exported below, so importing them from here keeps
+working.
```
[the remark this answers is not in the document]
mo-sameh: They are not imported outside of this module in alphabase, peptdeep and alphadia ?!

PR #446, opened 2026-09-02
[translate 3/3] Fix correctness and naming in the spectral library exports

mo-sameh:
### Goal

Fix the remaining defects the characterization tests exposed, now that the structure makes them one-line changes, and give the exports names that say what they produce.

### What's here

- **`rt_norm_pred` is accepted as a retention time.** peptdeep writes it alongside `rt_pred`, so a library whose `rt_pred` was dropped was rejected outright.
- **DIA-NN `Flags` marks one base peak per precursor row.** It grouped on `Precursor.Id`, which is not unique — `SpecLibBase.append` produces libraries holding the same precursor twice, and those shared a single flag. Two appended libraries: 8 flags for 16 precursor rows, now 16.
- **The SWATH fragment columns are typed.** They were `object`, and `FragmentCharge` held strings, so arithmetic on it concatenated. The written tsv and parquet are byte-identical; only the in-memory frames change.
- **`translate_to_tsv` notices a dead writer.** `multiprocessing=True` is the default, and that process dies before writing anything if the calling script has no `if __name__ == "__main__":` guard — the norm on macOS and Windows. The export printed its success message and returned over a 0-byte file.
- **`speclib_to_single_df` is now `speclib_to_swath_df`.** The good name already existed as a wrapper that has been missing its `return` since the module arrived from AlphaPeptDeep in 2022, so it only ever returned `None`. The real function takes the name; the old one stays as a deprecated alias. Nothing breaks — the wrapper's arguments are a subset, under the same names and defaults.

The module now reads `speclib_to_*_df` to build a frame, `translate_to_*` to write a file.


alphabase/spectral_library/translate_core.py
```diff
@@ -327,27 +322,27 @@ def fragment_table(  # noqa: PLR0913
...
+                [len(kept) for kept in frag_masses],
+            ),
             FragmentTableCols.FRAG_TYPE: frag_types,
-            FragmentTableCols.MZ: frag_masses,
-            FragmentTableCols.INTENSITY: frag_intensities,
-            FragmentTableCols.CHARGE: frag_charges,
-            FragmentTableCols.SERIES_NUMBER: frag_numbers,
+            FragmentTableCols.MZ: np.concatenate(frag_masses),
+            FragmentTableCols.INTENSITY: np.concatenate(frag_intensities),
+            FragmentTableCols.CHARGE: np.array(frag_charges, dtype=np.int64),
```
mschwoer, line 341: is `int64` required for charges?

alphabase/spectral_library/translate_core.py
```diff
@@ -327,27 +322,27 @@ def fragment_table(  # noqa: PLR0913
 
         infos = [_get_frag_info_from_column_name(column) for column in columns]
         types, losses, charges = zip(*infos) if infos else ((), (), ())
-        frag_types.append(types)
-        frag_losses.append(losses)
-        frag_charges.append(charges)
+        frag_types.extend(types)
```
mschwoer, line 325: we change from append to extend, is this intended?

alphabase/spectral_library/translate.py
```diff
@@ -73,7 +74,7 @@ def _precursors_to_swath_df(  # noqa: PLR0913
 ) -> pd.DataFrame:
     """Convert precursor and fragment frames to a SWATH transition list.
 
-    The dataframe-in form of :func:`speclib_to_single_df`, so that one batch of
+    The dataframe-in form of :func:`speclib_to_swath_df`, so that one batch of
```
GeorgWa, line 77: (I know this is not part of this PR)

The naming SWATH here is a bit confusing. I would maybe call it just "spectral library" and do a full rename in a separate PR.

GeorgWa (approved):
Looks good to me!

alphabase/spectral_library/translate_core.py
mo-sameh: Good catch, its not required in this part of the code. I switched to int 8

alphabase/spectral_library/translate_core.py
mo-sameh: Yes, the scalar columns need to be flat already, as we're not doing the explode any more and construct the dataframe in the "exploded" view directly.

alphabase/spectral_library/translate.py
mo-sameh: I had a hard time naming this too. I don't think plain "spectral library" works though because the DIA-NN parquet export is also a spectral library, so the name wouldn't tell the two apart.

PR #396, opened 2026-01-14
Fix spectronaut reader mapping

vbrennsteiner:
### Reasons for the PR:
- Reading Spectronaut reports from versions 20.0.250515.50606 and 20.0.250515.50606 failed with the `spectronaut_report` reader config due to missing columns (intensity, precursor_intensity). 
- `_pre_process()` tried to parse charge out of non-existent columns (`ModifiedSequence`)

### The issues
- Some columns that are present in these report files were not covered by the reader config (FG.PrecMz for `charge`, PG.ProteinNames for `genes`). 
- The `_preprocess()` method tried to parse charge from the mod_seq_column "ModifiedSequence", which does not exist in these reports. 

### The fixes
- Extend `spectronaut_report` configuration to parse additional columns correctly
- Add `EG.PrecursorId` to `mod_seq_columns` to get charge information in case `FG.PrecMz` is missing
- Modify `_pre_process()` to first check whether the charge column is present and only then parse it from the `precursor_id` column

PR #450, opened 2026-09-14
[fix] Spectronaut Reader - Column Mapping

lucas-diedrich:
Supersedes #396

Addresses the issues with the current `SpectronautReportReader` that fails for most reports with newer spectronaut reports (v20+). 

## Diagnostics 
See e.g. on ProteoBench output in [pbreadertest](https://github.com/lucas-diedrich/reader-continuous-integration.git)
```shell
pbreadertest --software Spectronaut
```

Key issues (as mostly described in #396)

1. The `SpectronautReportReader._pre_process` function expected that certain columns exist, which are not necessarily available (`precursor_id_column="EG.PrecursorId"`). 
2. The reader enforces (brittle) extraction of certain columns like charge and modified sequence, which are also available _as is_ in some spectronaut reports
3. Missing keys in config

Further:
- The `PSMReaderBase` class contains certain attributes that are _only_ used by the Spectronaut readers (`_precursor_id_columns`, `_precursor_id_column`)
- The `psm_reader.yaml` contains the key `precursor_id_columns` ONLY for the spectronaut readers, which conceptually collides with the `["column_mapping"]["precursor_id"]` key.

## Implementation
- Remove spectronaut-specific attributes + yaml-keys. 
- Add keys for later versions. Note that the key-mappings do currently not agree with #396


alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -673,22 +673,47 @@ diann: # 1.8.1
...
   reader_type: spectronaut_report
   rt_unit: minute
   fixed_C57: False
   column_mapping:
     'raw_name': 'R.FileName'
-    'rt': ['EG.ApexRT','EG.MeanApexRT']
-    'mobility': ['FG.ApexIonMobility']
-    'proteins': ['PG.ProteinNames','PG.ProteinGroups']
+    'rt':
+      - 'EG.ApexRT'
```
mschwoer, line 686: maybe add the respective versions as comments after each entry?

alphabase/psm_reader/dia_psm_reader.py
```diff
@@ -124,12 +126,85 @@ class SpectronautReportReader(ModifiedSequenceReader):
...
+            df,
+        )
+
+        # No need to overwrite charge or modified sequence if the report contains them
+        charge_available = PsmDfCols.CHARGE in get_column_mapping_for_df(
+            self.column_mapping, df
+        )
+        mod_seq_available = self.mod_seq_column is not None
+
+        # Return early if charge and modified sequence are available
```
mschwoer: redundant comment

alphabase/psm_reader/dia_psm_reader.py
```diff
@@ -124,12 +126,85 @@ class SpectronautReportReader(ModifiedSequenceReader):
...
+            return df
+
+        if not charge_available:
+            df[PsmDfCols.CHARGE] = self._extract_charge_from_precursor_column(
+                df[precursor_id_column]
+            )
+
+        if not mod_seq_available:
+            # `import_file()` sets mod_seq_column to None. Update to extracted column (standardized name)
+            # otherwise `_load_modifications()` looks up a `None` column
```
mschwoer: does this comment add value?

alphabase/psm_reader/dia_psm_reader.py
```diff
@@ -124,12 +126,85 @@ class SpectronautReportReader(ModifiedSequenceReader):
...
+        charge_available = PsmDfCols.CHARGE in get_column_mapping_for_df(
+            self.column_mapping, df
+        )
+        mod_seq_available = self.mod_seq_column is not None
+
+        # Return early if charge and modified sequence are available
+        if charge_available and mod_seq_available:
+            return df
+
+        # If precursor_id is missing, there is no way to extract the information
```
mschwoer: redundant comment (to code and to how life works in general :-D )

alphabase/psm_reader/psm_reader.py
```diff
@@ -153,11 +153,6 @@ def __init__(  # noqa: PLR0913 # too many arguments
         self._fdr_threshold = fdr
         self._keep_decoy = keep_decoy
 
-        self._precursor_id_columns = psm_reader_yaml[self._reader_type].get(
```
mschwoer, line 156: this changes the behaviour for all classes derived from here, please double-check for side effects

alphabase/psm_reader/psm_reader.py
```diff
@@ -238,9 +233,6 @@ def import_file(self, _file: str) -> pd.DataFrame:
             self.mod_seq_column = self._get_actual_column(
                 self._mod_seq_columns, origin_df
             )
-            self._precursor_id_column = self._get_actual_column(
```
mschwoer, line 241: this changes the behaviour for all classes derived from here, please double-check for side effects

alphabase/constants/const_files/psm_reader.yaml
```diff
@@ -673,22 +673,47 @@ diann: # 1.8.1
...
+      - 'FG.LabeledSequence'
+      - 'FG.XICDBID'
+    'precursor_intensity':
+      - 'FG.Quantity'
+    'precursor_mz': 'FG.PrecMz'
+    'sequence':
+      - "PEP.StrippedSequence"
+    'fdr':
+      - "FG.Qvalue"
+    'protein_fdr':
```
mschwoer, line 712: is there like a "minimal set" of Spectronaut fields that come always, regarding how the report is configured? if not, or if we need more than the minimal set: should we somewhere document what we need?

alphabase/psm_reader/dia_psm_reader.py
lucas-diedrich: When I implemented this, I had forgotten that the base class has this attribute and that it's important for the downstream standardization/should not be None. but the comment is incomprehensible.

alphabase/psm_reader/psm_reader.py
lucas-diedrich: `_precursor_id_columns` does not show up in the code base anymore (search/lookup) and was only referenced in the spectronaut report at a single occasion

alphabase/psm_reader/psm_reader.py
lucas-diedrich: `_precursor_id_column` does not show up in the code base anymore (search/lookup) 
and it's a private attribute

alphabase/psm_reader/dia_psm_reader.py
lucas-diedrich: 🙈

alphabase/constants/const_files/psm_reader.yaml
lucas-diedrich: should we do this on a reader-level? i.e. `PsmDfCols`? One could for example replace the `PsmDfCols` class (which currently only standardizes the access between columns but does not validate that `psm_reader_yaml` is conformant to this standardization) with a pydantic class with required + optional attributes.

mschwoer: that would be the technical check, but I am more thinking about telling users "if you want to use apt with spectronaut data, make sure your report contains X Y Z) ..

lucas-diedrich: This is basically the `alphapepttools.io.reader_columns.FEATURE_LEVEL_CONFIG` *
alphapepttools just needs a sample_id column, feature_id column, and a  intensity, it does not _necessarily_ care about standardization

I guess we could expose this in alphapepttools to users + the `psm_reader_yaml` to include standardized + default search engine names. 

But alphabase would like to have additional columns (ideally precursor/protein FDR, modification columns, retention time ...). I am currently not completely sure what we consider the "standard" functionality of alphabase.

alphabase/psm_reader/psm_reader.py
lucas-diedrich: @mschwoer is this resolved?

alphabase/constants/const_files/psm_reader.yaml
lucas-diedrich: I can do this for versions 19-21 (proteobench versions)

lucas-diedrich:
Should I add an integration test with a report that failed before? @mschwoer

mschwoer:
@lucas-diedrich if you have the integration test data at hand, why not

mschwoer (approved):

lucas-diedrich:
Added integration tests and validate that they fail on current main

alphabase/constants/const_files/psm_reader.yaml
lucas-diedrich: this is version independent since v18

PR #451, opened 2026-09-14
[fix] Datatype inference in `psm_reader` _load_file

lucas-diedrich:
Fixes the issue that numeric columns with missing values are converted to object dtypes in the `psm_reader` module

## Diagnostics 
Surfaced in real-world data: See e.g. on ProteoBench output in [pbreadertest](https://github.com/lucas-diedrich/reader-continuous-integration.git)
```shell
pbreadertest --software Spectronaut
```

### Root cause 
The current argument `keep_default_na=False` in `pd.read_csv` interprets all NA values as strings. In numeric PSM table columns this coerces the intensity columns to object dtype (which is incorrect). 

`pd.read_csv(..., keep_default_na=False)` is used because NA is a special value in the modification mapping columns, indicating that the modification could not be mapped to a known modification (psm reader drops the object rows with NAs). 

This PR introduces a utility function `_coerce_object_nan_to_empty_string` to allow us to use the default NA-inference of `pd.read_csv` while keeping the contract that missing values in object columns are encoded as empty strings (`''`)


alphabase/psm_reader/psm_reader.py
```diff
@@ -265,18 +265,32 @@ def _load_file(self, filename: Union[str, Path, io.StringIO]) -> pd.DataFrame:
...
+            - Missing values in object columns are encoded as empty strings.
+            - Missing values in numeric columns are encoded as np.nan
+
         """
         if isinstance(filename, io.StringIO):
             sep = _get_delimiter(filename)
-            return pd.read_csv(filename, sep=sep, keep_default_na=False)
+            df = pd.read_csv(filename, sep=sep, keep_default_na=True)
+            # NAs have a special meaning as they indicate unresolved modifications
+            # Use empty strings in object/string columns to indicate missing values
```
mschwoer: please move this comment to `_coerce_object_nan_to_empty_string`

alphabase/utils.py
```diff
@@ -70,3 +70,53 @@ def _get_delimiter(file_path: str) -> str:
         return ","
     else:
         return "\t"
+
+
+def _coerce_object_nan_to_empty_string(df: pd.DataFrame) -> pd.DataFrame:
```
mschwoer: this is a quite specific name, leaks this logic into the calling code .. could it be smth `_sanitize_dataframe()` ? 
You have every right to push back -- say a word and I'll step back from this comment.

:-D

alphabase/utils.py
```diff
@@ -70,3 +70,53 @@ def _get_delimiter(file_path: str) -> str:
...
+    -------
+    Copy of dataframe with coerced columns
+
+    Example
+    -------
+    .. code-block:: python
+
+        df["numeric_column"]
+        >   0    1.0
+            1    2.0
```
mschwoer: (nit) a two-row example df would buy us 3 LoC :-)

mschwoer (approved):

alphabase/utils.py
lucas-diedrich: claude actually used the name `restore_empty_strings`, so joke's on you

alphabase/utils.py
lucas-diedrich: wanted to bloat our code base a little more

alphabase/utils.py
lucas-diedrich: Renamed to `_sanitize_missing_values`

PR #452, opened 2026-09-14
[fix] Adapt search engine-specific PSM readers to new convention in `_load_file`

lucas-diedrich:
Adapt individual readers to the convention introduced in #451 

I think we can drop the reader-specific `_load_file` calls most of the time as the base reader basically covers these use cases already.


alphabase/psm_reader/maxquant_reader.py
```diff
@@ -240,13 +240,13 @@ def __init__(  # noqa: PLR0913, D417 # too many arguments in function definition
...
         """MaxQuant-specific preprocessing of output data."""
         df = df[~pd.isna(df["Retention time"])]
-        df.fillna("", inplace=True)
 
         # remove MBR PSMs as they are currently not supported and will crash import
         mapped_columns = get_column_mapping_for_df(self.column_mapping, df)
         if PsmDfCols.SCAN_NUM in mapped_columns:
             scan_num_col = mapped_columns[PsmDfCols.SCAN_NUM]
-            no_ms2_mask = df[scan_num_col] == ""
+            # MBR PSMs carry no scan number: blank in a text column, NaN in a numeric one
```
mschwoer: isn't this solved by  `_coerce_object_nan_to_empty_string`? https://github.com/MannLabs/alphabase/pull/451/changes

GeorgWa (approved):
Looks good!

